PR #12, opened 2017-02-06
Exclude .DS_Store generated by Mac OSX

marcocucinato:
Exclude from Git the .DS_Store directory, generated by Mac OSX.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -87,3 +87,6 @@ ENV/
 
 # JetBrains
 .idea/
+
+# Mac stuff
+.DS_Store


adobeDan:
@marcocucinato, thanks for the suggestion!  Since this is a cross-platform app, our convention is that developers on specific platforms should have a global git ignore file on their machine that excludes their platform-specific artifacts.  (See [this article](https://help.github.com/articles/ignoring-files/) for details on how to do that.  There is a [recommended list of Mac OS ignores](https://github.com/github/gitignore/blob/afbff9027d02ccfc680e031f6c295f79ad61662d/Global/macOS.gitignore) maintained by GitHub in their [github ignore repository](https://github.com/github/gitignore).)

If you believe it's important that we should have all the various platform ignores as part of our repo .gitignore, please make that case by [opening an issue](https://github.com/adobe-apiplatform/user-sync.py/issues/new) -- feel free to attach an appropriate PR -- and we will consider it.

PR #15, opened 2017-02-07
Add error for invalid file and console log levels

jamesg123:


jamesg123:
@adobeDan agreed that is a better approach. I have gone with a default of debug for error level (i.e. everything) so that the user doesn't miss errors if they specify an invalid option.

user_sync/app.py
```diff
@@ -105,11 +105,16 @@ def init_log(logging_config):
     }
     
     console_log_level = level_lookup.get(options['console_log_level'])
-    if (console_log_level != None):
-        console_log_handler.setLevel(console_log_level)
-    
+    if (console_log_level == None):
+        console_log_level = logging.DEBUG
```
phil-levy: After a security review with our audit team, they requested that the default logging level be Info rather than debug.  Affects message a couple of lines down also, and file log level also.

user_sync/app.py
```diff
@@ -105,11 +105,16 @@ def init_log(logging_config):
...
+        logger.log(logging.INFO, 'Unknown console log level: %s setting to debug' % options['console_log_level'])
+    console_log_handler.setLevel(console_log_level)
+
     if options['log_to_file'] == True:
-        file_log_level = level_lookup.get(options['file_log_level'], logging.NOTSET)
+        file_log_level = level_lookup.get(options['file_log_level'])
+        if (file_log_level == None):
+            file_log_level = logging.DEBUG
+            logger.log(logging.INFO, 'Unknown file log level: %s setting to debug' % options['file_log_level'])
```
phil-levy: This line (and the corresponding one above seem problematic.  Either we are going to be emitting this message about unknown log level even if no log level has been specified, or this code is not running to set the default if there is no specification in the file.  Seems like this line (and the one above) should only emit if there is an invalid value specified.

Also, this should be a warning, not info.

jamesg123: @phil-levy - I have updated the logger.log to warning.

I think it is ok as the file log level is defaulted with this line in the init:

builder.set_string_value('file_log_level', 'debug')

* If file_log_level is not specified, then we will use the default specified in the init and we won’t output the message
* If file log_level is set to an invalid value e.g. TEST, then none will be returned from the .get, we will set the level to debug and output the message
* If file log_level is set to a valid value, then we don’t go into the if, we use the valid file_log_level.

This would be an issue however with the console_log_level as the default set for that in the init is None, therefore if it was missing from the config file you would get the error message and have the value defaulted. Changing the default value to info (in response to your other comment) will address this I believe.

jamesg123:
@phil-levy I have updated: 
*Both the logs about invalid file/console settings to be warning level (rather than info) and I have updated the text to say the default is info.
* I have updated the defaults for file/console levels to info. Note I have assumed that this default should also be used when the settings are not present in the file and when the value specified is invalid. Please correct me if this assumption is incorrect.
* I also noticed that the log message regarding invalid file log level was not written to the log file as the file hadn't been initialised at that point. I have changed this now so that the error is flagged and reported after the log file is setup.

PR #20, opened 2017-02-15
Report error for federated users with blank country code and no defau…

jamesg123:
…lt also update unit tests

### DIFF
--- a/user_sync/rules.py
+++ b/user_sync/rules.py
@@ -352,11 +352,14 @@ class RuleProcessor(object):
         country = directory_user['country']
         if not country:
             country = options['default_country_code']
-        if identity_type == user_sync.identity_type.ENTERPRISE_IDENTITY_TYPE and not country:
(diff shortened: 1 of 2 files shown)


user_sync/rules.py
```diff
@@ -352,11 +352,14 @@ def add_dashboard_user(self, user_key, dashboard_connectors):
         country = directory_user['country']
         if not country:
             country = options['default_country_code']
-        if identity_type == user_sync.identity_type.ENTERPRISE_IDENTITY_TYPE and not country:
-            # Enterprise users are allowed to have undefined country
-            country = 'UD'
-        if (country != None):
-            attributes['country'] = country    
+        if not country:
+            if identity_type == user_sync.identity_type.ENTERPRISE_IDENTITY_TYPE:
```
phil-levy: What happens here for Adobe ID?  Adobe Id does not require a country either.

PR #28, opened 2017-02-27
Features/max deletions and max missing users

jamesg123:


jamesg123:
Closing this to make adjustment to config.py

PR #31, opened 2017-02-28
Complex Mapping feature (LIGHTLY TESTED, NEEDS REVIEW)

Metasemi:


user_sync/config.py
```diff
@@ -312,6 +307,40 @@ def get_rule_options(self):
...
+                            # [TODO morr 2017-02-27]: Do we really need to pre-create extended dashboard groups here? Or
+                            # could it be done on the fly, when they're encountered in values returned from hook code? If
+                            # the latter, we could do the customer a big favor by not requiring them to be declared in the
+                            # extension config.
+                            #
+                            # This should be revisited once the Complex Mapping feature as a whole is working and has been
+                            # thoroughly tested.
+                            #
+                            for extended_dashboard_group in extension_config.get_list('extended_dashboard_groups'):
+                                group = self.create_dashboard_group(extended_dashboard_group)
```
adobeDan: I can see from the code later on that `create_dashboard_group` ignores the group if it already exists as a dashboard group.  This means you are guarding against people specifying groups in their `extended_dashboard_groups` that are already groups in the mapping, which seems like it would be a common (and forgiveable) mistake.  A comment to that effect would be useful here (and also in the docs).

user_sync/connector/directory_csv.py
```diff
@@ -37,12 +37,16 @@ def connector_initialize(options):
...
-def connector_load_users_and_groups(state, groups):
+def connector_load_users_and_groups(state, groups, extended_attributes):
     '''
     :type state: CSVDirectoryConnector
     :type groups: list(str)
+    :type extended_attributes: list(str)
     :rtype (bool, iterable(dict))
     '''
+    if (extended_attributes is not None):
+        self.logger.warning("CSV directory connector doesn't support extended_attributes; ignored")
```
adobeDan: Really?  Seems like it would be really really easy for the CSV connector to support extended attributes, since you could just add them as column headers.  And this seems like it would be a really, really handy way to test extended logic on just a few people...

user_sync/connector/directory_ldap.py
```diff
@@ -269,35 +272,55 @@ def iter_users(self, users_filter):
...
             
             user = user_sync.connector.helper.create_blank_user()
             user['email'] = email
+            source_attributes['email'] = email
                 
             username, last_attribute_name = self.user_username_formatter.generate_value(record)
             if (username == None and last_attribute_name != None):
                 self.logger.info('No username attribute: %s for dn: %s', last_attribute_name, dn)    
             user['username'] = username if username != None else email
+            if (username != None):
```
adobeDan: Why is this conditionalized?  Seems like you should be setting the `source_attributes['username']` to be `user['username']` so that it always has a value.  Not setting it at all is likely to lead to assertions when it's accessed in the user extension code.

user_sync/connector/directory_ldap.py
```diff
@@ -269,35 +272,55 @@ def iter_users(self, users_filter):
...
             if (domain != None):
                 user['domain'] = domain
+                source_attributes['domain'] = domain
             elif (last_attribute_name != None):
                 self.logger.info('No domain attribute: %s for dn: %s', last_attribute_name, dn)    
                                                 
             given_name_value = LDAPValueFormatter.get_attribute_value(record, 'givenName')
             if (given_name_value != None):   
                 user['firstname'] = given_name_value
+                source_attributes['givenName'] = given_name_value
```
adobeDan: Why not make the keys of `source_attributes` match the keys of the source customer record (i.e., `firstname` and `lastname` and `country`) as constructed here?  I don't think we want to force the customer to know which particular alias for `gn` and `sn` we are using in our code.  The actual source attribute in the directory could be any one of the LDAP aliases for `gn` and `sn`, so it's not like we are guaranteed to be using the actual value in the source directory. So as long as we are picking `firstname` and `lastname` as the canonical values we use on the Adobe side I think we should just use them on the customer side in the extensions as well.  After all, if the extension wants to set those value it's going to have to know those names.

user_sync/config.py
```diff
@@ -312,6 +307,40 @@ def get_rule_options(self):
...
             self.logger.warning("Assuming the identity type for users is: %s", new_account_type)
+
+        after_mapping_hook = None
+        extended_attributes = None
+        extensions_config = self.main_config.get_list_config('extensions', True)
+        if (extensions_config != None):
+            for extension_config in extensions_config.iter_dict_configs():
+                context = extension_config.get_string('context')
+                if context == 'per-user':
+                    if (after_mapping_hook == None):
```
phil-levy: Is this test needed/correct?  Looks like after_mapping_hook set to None a few lines up.

user_sync/connector/directory_ldap.py
```diff
@@ -269,35 +272,55 @@ def iter_users(self, users_filter):
...
             sn_value = LDAPValueFormatter.get_attribute_value(record, 'sn')
             if sn_value != None:
                 user['lastname'] = sn_value
+                source_attributes['sn'] = sn_value
```
adobeDan: See comment at line 299.

user_sync/connector/directory_ldap.py
```diff
@@ -269,35 +272,55 @@ def iter_users(self, users_filter):
...
             c_value = LDAPValueFormatter.get_attribute_value(record, 'c')
             if c_value != None:
                 user['country'] = c_value
+                source_attributes['c'] = c_value
```
adobeDan: See comment at line 299.

user_sync/config.py
Metasemi: This is testing for multiple per-user extensions; see the matching else.

user_sync/connector/directory_ldap.py
```diff
@@ -269,35 +272,55 @@ def iter_users(self, users_filter):
...
             uid = LDAPValueFormatter.get_attribute_value(record, 'uid')
             if (uid != None):
                 user['uid'] = uid
-            
+                source_attributes['uid'] = uid
+
+            if extended_attributes is not None:
+                for extended_attribute in extended_attributes:
+                    extended_attribute_value = LDAPValueFormatter.get_attribute_value(record, extended_attribute)
+                    if (extended_attribute_value is not None):
```
adobeDan: This looks to be brittle if any of the base attributes we look for are also specified in the extended attributes, because we will overwrite the base attributes (which have special logic applied to them) with the raw value.  Are you checking to make sure there is no overlap?

Also this is another argument for the change I suggested at line 299, if we make the source_attributes for the Adobe-relevant values have the Adobe-relevant names, then if the code really wants the value of the `gn` attribute in the source it can specify `gn` as an extended attribute and get it separately direct from the source.

user_sync/rules.py
```diff
@@ -126,6 +142,7 @@ def read_desired_user_groups(self, mappings, directory_connector):
         directory_group_filter = options['directory_group_filter']
         if (directory_group_filter != None):
             directory_group_filter = set(directory_group_filter)
+        extended_attributes = options.get('extended_attributes')
```
adobeDan: See comment above at line 317.  We need to make sure these don't overlap with the base attributes.

user_sync/rules.py
```diff
@@ -152,14 +169,45 @@ def read_desired_user_groups(self, mappings, directory_connector):
...
-                        organization_info.add_desired_group_for(user_key, dashboard_group.group_name)
-    
-        self.logger.info('Total directory users after filtering: %d', len(filtered_directory_user_by_user_key))        
+                        self.after_mapping_hook_scope['target_groups'].add(self.make_dashboard_group_qualified_name(dashboard_group.group_name, dashboard_group.organization_name))
+
+            # only if there actually is hook code: set up rest of hook scope, invoke hook, update user attributes
+            if (options['after_mapping_hook'] is not None):
+                self.after_mapping_hook_scope['source_attributes'] = directory_user['source_attributes'].copy()
+
+                # [TODO morr 2017-02-26]: Is there a more robust way to assemble the target attributes?
```
adobeDan: What exactly are you worried about?  That there will be `directory_user` attributes that aren't on the Adobe side?  Then just use a comprehension constructor
```
{k: directory_user[k] for k in ('username', 'email', 'country', 'firstname', 'lastname')}
```

user_sync/config.py
phil-levy: Got it.  Missed the loop.

user_sync/rules.py
```diff
@@ -152,14 +169,45 @@ def read_desired_user_groups(self, mappings, directory_connector):
...
+                self.after_mapping_hook_scope['target_attributes'] = directory_user.copy()
+                self.after_mapping_hook_scope['target_attributes'].pop('groups', None)
+                self.after_mapping_hook_scope['target_attributes'].pop('identitytype', None)
+                self.after_mapping_hook_scope['target_attributes'].pop('source_attributes', None)
+
+                # invoke the customer's hook code
+                # [TODO morr 2017-02-27: Not putting any guardrails around the customer's code, which is treated as
```
adobeDan: You're giving them a confined scope; there's nothing really they can do except crash us or hang us.  And there's no good way to guard against that :).

user_sync/rules.py
```diff
@@ -152,14 +169,45 @@ def read_desired_user_groups(self, mappings, directory_connector):
...
+            for target_group_qualified_name in self.after_mapping_hook_scope['target_groups']:
+                target_group_name, target_organization_name = self.parse_dashboard_group_qualified_name(target_group_qualified_name)
+                target_group = Group.get_dashboard_group(target_group_name, target_organization_name)
+                if (target_group is not None):
+                    organization_info = self.get_organization_info(target_organization_name)
+                    organization_info.add_desired_group_for(user_key, target_group_name)
+                else:
+                    self.logger.error('Target dashboard group %s is not known; ignored', target_group_qualified_name)
+
+        self.logger.info('Total directory users after filtering: %d', len(filtered_directory_user_by_user_key))
```
adobeDan: Is there some corresponding log of Total directory users before filtering?  If not, why not, and if so, why isn't this statement next to it?

user_sync/rules.py
```diff
@@ -617,8 +679,16 @@ def write_remove_list(self, file_path, dashboard_users):
                 writer.writerow({'user': username, 'domain': domain})
                 total_users += 1
         self.logger.info('Total users in remove list: %d', total_users)
-                
-        
+
+    def log_after_mapping_hook_scope(self, before_call):
```
adobeDan: Ouch!  A boolean variable named `before_call`, so the call has `True` or `False` in it?  That's really opaque!  Either use keyword variables `before_call=False`, `after_call=False` and have the caller specify one of them (and err if caller doesn't), or have a variable `context` which has a string value which must be `before_call` or `after_call` (and is an error otherwise).

user_sync/rules.py
```diff
@@ -152,14 +169,45 @@ def read_desired_user_groups(self, mappings, directory_connector):
...
+                # [TODO morr 2017-02-26]: Is there a more robust way to assemble the target attributes?
+                self.after_mapping_hook_scope['target_attributes'] = directory_user.copy()
+                self.after_mapping_hook_scope['target_attributes'].pop('groups', None)
+                self.after_mapping_hook_scope['target_attributes'].pop('identitytype', None)
+                self.after_mapping_hook_scope['target_attributes'].pop('source_attributes', None)
+
+                # invoke the customer's hook code
+                # [TODO morr 2017-02-27: Not putting any guardrails around the customer's code, which is treated as
+                # friendly. Is there anything we should be doing?
+                self.log_after_mapping_hook_scope(True)
```
adobeDan: Ouch!  How opaque is this!  See comment at line 683.

user_sync/rules.py
```diff
@@ -617,8 +679,16 @@ def write_remove_list(self, file_path, dashboard_users):
                 writer.writerow({'user': username, 'domain': domain})
                 total_users += 1
         self.logger.info('Total users in remove list: %d', total_users)
-                
-        
+
+    def log_after_mapping_hook_scope(self, before_call):
+        when = 'before' if before_call else 'after'
+        if (before_call):
+            self.logger.debug('.')
```
adobeDan: what is this about?  are you trying to get a separator line in the log?  I bet we could be a lot clearer than a simple period.

user_sync/connector/directory_csv.py
phil-levy: Doesn't really make sense for csv to have extended attrs because all csv attrs are always included in the data.  No need to ask it to include more.

user_sync/connector/directory_ldap.py
phil-levy: May be moot because I think it is always set in user results now, but seems like leaving it out when there is no value is reasonable vs having it present with a None value.

user_sync/connector/directory_csv.py
adobeDan: Right, that's exactly my point 👍 .  We are warning the user that the CSV connector doesn't support extended attributes but in fact I believe the CSV connector *already* supports extended attributes without our doing any coding!

user_sync/rules.py
phil-levy: What about exceptions flying out?

user_sync/connector/directory_ldap.py
adobeDan: I disagree, because having no value results in an assertion when you access the attribute source_attributes['username'].  If you are trying to make it clear that there was no source attribute 'username' then return None for its value, don't force the user into awkward constructions like source_attributes.get('username') just for some attributes and not others.

user_sync/rules.py
adobeDan: Right, that's the crash case.  I don't think we should guard against that, because we can't continue, and it's pointless for us to catch, log, and rethrow.

user_sync/connector/directory_csv.py
phil-levy: So I think the action here is to silently ignore the extended attrs list.  Just like with AD, if there is no such attr, it just won't show up.

user_sync/connector/directory_ldap.py
phil-levy: I'm convinced.  Have value be None.

user_sync/connector/directory_csv.py
adobeDan: Right, so Phil and I agree: just remove this test.  The CSV directory connector *does* support extended attributes, as far as we know.  Have you verified otherwise, Mike?

user_sync/config.py
```diff
@@ -198,15 +198,10 @@ def get_directory_groups(self):
...
             dashboard_groups_config = item.get_list_config('dashboard_groups')
             for dashboard_group in dashboard_groups_config.iter_values(types.StringTypes):
-                parts = dashboard_group.split(GROUP_NAME_DELIMITER)
-                group_name = parts.pop()
-                organization_name = GROUP_NAME_DELIMITER.join(parts)
-                if (len(organization_name) == 0):
-                    organization_name = user_sync.rules.OWNING_ORGANIZATION_NAME
-                if (len(group_name) == 0):
+                group = self.create_dashboard_group(dashboard_group)
+                if (group is None):
```
Metasemi: Yes, following the style of the codebase...

user_sync/config.py
```diff
@@ -312,6 +307,40 @@ def get_rule_options(self):
         if (new_account_type == None):
             new_account_type = user_sync.identity_type.ENTERPRISE_IDENTITY_TYPE
             self.logger.warning("Assuming the identity type for users is: %s", new_account_type)
+
+        after_mapping_hook = None
+        extended_attributes = None
+        extensions_config = self.main_config.get_list_config('extensions', True)
+        if (extensions_config != None):
```
Metasemi: I used the equality when in the midst of a block of code that was already using it; otherwise I did what you're calling the right way. Should be addressed globally.

user_sync/config.py
```diff
@@ -312,6 +307,40 @@ def get_rule_options(self):
...
+            for extension_config in extensions_config.iter_dict_configs():
+                context = extension_config.get_string('context')
+                if context == 'per-user':
+                    if (after_mapping_hook == None):
+                        after_mapping_hook_text = extension_config.get_string('after_mapping_hook')
+                        if (after_mapping_hook_text is not None):
```
Metasemi: Leaving this for now, but agree with you. Will fix if I have time.

user_sync/config.py
Metasemi: Actually create_dashboard_group did *not* behave that way. But it does now :)

user_sync/connector/directory_csv.py
Metasemi: Great, done.

user_sync/connector/directory_ldap.py
Metasemi: The before state of target_attributes is exactly what you're asking for, isn't it?

user_sync/connector/directory_ldap.py
Metasemi: Great catch. Fixed to discard the colliders silently.

user_sync/rules.py
Metasemi: Not worried. Choices were keep everything, keep the known-to-be-real, discard the known-to-be-plumbing. I did discard the known-to-be-plumbing and still think it's best. Removing the comment.

user_sync/rules.py
Metasemi: Agreed, removing the comment. Similarly we don't catch errors compiling the hook code, and the resulting behavior is excellent.

user_sync/rules.py
Metasemi: Leaving alone for now.

user_sync/rules.py
Metasemi: This was not changed by me, but fwiw this is not confusing in the actual log, modulo the fact that the term "filtering" is used in multiple ways - the --users way and the all_users_filter way. (This is talking about the former.) Anyway I'm not doing anything here.

user_sync/rules.py
Metasemi: Agree with your complaint, though I can't say I'm crazy about either proposed fix. Leaving alone for now.

user_sync/rules.py
Metasemi: It's pretty great the way it is. What would you suggest? :)

user_sync/connector/directory_ldap.py
Metasemi: You're actually making two different points here. (1) You want username to be overridden with email if not present. I definitely disagree with this. It loses information, and what you're looking for already exists in target_attributes. (2) You want all attributes we know about to set to None if not present. I reacted negatively to this at first to, but as I try to figure out why, I think it's because it makes the log harder to scan. But maybe that's not a good enough reason. (Yeah, yeah, I know, you're going to say that's about us, not about them. But actually I think they'll be living in the logs too...)

PR #38, opened 2017-03-01
Fix #26: use correct username and domain for dashboard users.

### LINKED ISSUE #26: Sync fails to recognize federated users with username-based login and keeps recreating them
If a federated domain is set up for username based login (rather than email based), then user sync does not seem to recognize users.  If you run the same sync repeatedly, it creates the same users and adds them to the same PLCs on  every run.

adobeDan:
For some reason the username and domain returned by the dashboard for a user were being ignored when constructing the `user_key`.  For federatedID users where the username is not the email, this caused the user key to differ from the one constructed for directory users.

### DIFF
--- a/user_sync/rules.py
+++ b/user_sync/rules.py
@@ -663,7 +663,7 @@ class RuleProcessor(object):
         '''
         :type dashboard_user: dict
         '''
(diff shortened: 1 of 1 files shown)


adobeDan:
Getting this in quickly in case Chester can build it.

PR #60, opened 2017-03-10
Pre-release testing merge from v1

adobeDan:
I want to merge the latest work from v1 into master for two reasons:

* so we can post the current 1.1.1 docs in their new format to GitHub Pages
* so we can start people testing a number of the new bug fixes that have been in v1

No version number bump in this pull request yet; I will introduce one before we merge.


adobeDan:
On second thought I am withdrawing this request.  I will do a contained request just for the doc changes (that match the 1.1.1 release) and separately we can figure out the pre-release strategy.

PR #61, opened 2017-03-10
Bump the version number to prep for the 1.1.2 release.

adobeDan:


adobeDan:
@phil-levy Can you do a quick review and approve so I can get this into v1?  I'm assuming Monday's release will be 1.1.2, but maybe you want it to be 1.2 since we will be adding adobeID support which might be considered more than a minor bug fix?

PR #62, opened 2017-03-10
Bring v1.1.1 docs to master.

adobeDan:
* This replaces the entire docs/ tree except for the "User Sync Guide.pdf" file.
* This merges in the required .gitignore changes for Jekyll.

PR #63, opened 2017-03-10
Bring v1.1.1 docs to master for publish to GitHub Pages.

adobeDan:
* This replaces the entire docs/ tree except for the "User Sync Guide.pdf" file.
* This merges in the required .gitignore changes for Jekyll.


adobeDan:
Unbelievably, I merged these changes, which came from v1, back into v1 (see pull #62)!  So now they need to go into master for real.  I think I will self-approve this, since it's already been approved earlier under a different name!

PR #73, opened 2017-03-14
added support for group_type in dashboard add_groups/remove_groups methods

ianmak:


adobeDan:
@ianmak I don't understand the purpose of this pull request.  Is this going to be the beginning of a bigger pull request around adding support for group types in general?  I don't think we need this change until we have a strategy for what we are going to do with group types generally.  When you create pull requests, it's helpful when you create them to give all the relevant context/intent/design notes etc. in the commentary.

ianmak:
Basically the purpose of this change was in preparation of adding support for user groups. Should it go in with the user groups change? I understand you were saying there's debate about whether it's even needed now...

adobeDan:
Yeah, I don't want to bring this stuff in except as part of an overall change to specify group types, and I don't think we're ready to make that overall change for this release.  So I'll close this PR and you should keep all the work you've done on this in your fork so it doesn't get lost.  If you want, feel free to submit a PR with all that work into a feature branch in this repo.

PR #75, opened 2017-03-14
updated documentation to include instructions for "--users mapped" parameter

ianmak:
basically updated main README, user manual, and success guide to make
it clear to the user that `--users mapped` is shorthand for including all
directory groups in the sync.

### DIFF
--- a/docs/success-guide/command_line_options.md
+++ b/docs/success-guide/command_line_options.md
@@ -15,7 +15,7 @@ The command line for user sync selects the set of users to be processed, specifi
 | ------------- |:-------------| 
 |   `--users all` |    All users listed in the directory are included.  |
 |   `--users group "g1,g2,g3"`  |    The named directory groups are used to form the user selection. <br>Users that are members of any of the groups are included.  |
-|   `--users mapped`  |    The same as --users group [...], but includes all directory groups defined in the configuration file. |
+|   `--users mapped`  |    The same as `--users group g1,g2,g3,...`, where `g1,g2,g3,...` are all the directory groups specified in the configuration file group mapping.|
 |   `--users file f`  |    The file f is read to form the selected set of users.  The LDAP directory is not used in this case. |
 |   `--user-filter pattern`    |  Can be combined with the above options to further filter and reduce the user selection. <br>`pattern` is a string in Python regular expression format.  <br>The user name must match the pattern in order to be included.  <br>Writing patterns can be somewhat of an art.  See examples below or refer to the Python documentation [here](https://docs.python.org/2/library/re.html). |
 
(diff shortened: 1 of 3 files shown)


docs/success-guide/command_line_options.md
```diff
@@ -15,6 +15,7 @@ The command line for user sync selects the set of users to be processed, specifi
 | ------------- |:-------------| 
 |   `--users all` |    All users listed in the directory are included.  |
 |   `--users group "g1,g2,g3"`  |    The named directory groups are used to form the user selection. <br>Users that are members of any of the groups are included.  |
+|   `--users mapped`  |    The same as --users group [...], but includes all directory groups defined in the configuration file. |
```
adobeDan: The directory groups are not "defined" in the configuration file, and this construction makes it look like there might be an optional argument.  I suggest "The same as `--users group g1,g2,g3,...` where `g1,g2,g3,...` are all the directory groups specified in the configuration file group mapping."

docs/user-manual/index.md
```diff
@@ -782,7 +782,7 @@ specific behavior in various situations.
...
-| `--users` `all`<br />`--users` `file` _input_path_<br />`--users` `group` _grp1,grp2_ | Specify the users to be selected for sync. The default is `all` meaning all users found in the directory. Specifying `file` means to take input user specifications from the CSV file named by the argument.  Specifying `group` interprets the argument as a comma-separated list of groups in the enterprise directory, and only users in those groups are selected. |
+| `--users` `all`<br />`--users` `file` _input_path_<br />`--users` `mapped`<br />`--users` `group` _grp1,grp2_ | Specify the users to be selected for sync. The default is `all` meaning all users found in the directory. Specifying `file` means to take input user specifications from the CSV file named by the argument. Specifying `mapped` is the same as specifying `group` with all groups listed in the configuration file. Specifying `group` interprets the argument as a comma-separated list of groups in the enterprise directory, and only users in those groups are selected. |
```
adobeDan: move `mapped` *after* `group` (both on the left side and the right side), because the definition of mapped depends on the definition of group.  I also suggest that rather than the phrase "listed in the configuration file" you say "listed in the group mapping in the configuration file".  It means the same thing, but makes it clear where to find them.

docs/user-manual/index.md
```diff
@@ -914,6 +914,15 @@ configuration or user group management.
 ./user-sync –c user-sync-config.yml --users groups "group1, group2, group3"
 ```
 
+#### Sync only users in given groups
+
+This action is the same as specifying --users groups "...", but adds all
```
adobeDan: Be sure to use backticks around `--users groups`.  Same comment as above, please reword this to make clear that there is no "..." argument.  Also no but.  Also there is only one configuration file where the group mapping exists.

adobeDan:
@ianmak - Almost there!  See new comments!

PR #77, opened 2017-03-14
 "Product License Configuration" -> "product configuration"

jabogart:


jabogart:
The only remaining reference to PLC is in a graphic image.

PR #81, opened 2017-03-15
added support for delete command line options

ianmak:
- added support for —delete-nonexistent-users, —generate-delete-list,
—delete-list
- updated documentation


user_sync/app.py
```diff
@@ -75,6 +75,15 @@ def process_args():
     parser.add_argument('-d', '--remove-list',
                         help='specifies the file containing the list of users to be removed. Users on this list are removeFromOrg\'ed on the Adobe side.',
                         metavar='input_path', dest='remove_list_input_path')
+    parser.add_argument('--delete-nonexistent-users',
```
phil-levy: Are you adding --remove-entitlements in a separate submission?

user_sync/app.py
```diff
@@ -75,6 +75,15 @@ def process_args():
...
+                        help='Causes the user sync tool to delete users that exist on the Adobe side if they are not in the customer side AD. If the account is an Adobe account, it is remove from org\'d, but not deleted.',
+                        action='store_true', dest='delete_nonexistent_users')
+    parser.add_argument('--generate-delete-list',
```
phil-levy: --generate-delete-list actually isn't in the spec because it does the same thing as --generate-remove-list.  Could be a reasonable thing to add.  Thoughts?

user_sync/app.py
ianmak: ah yeah I thought I'd add it in a different submission, as I just wanted to make sure the change was headed in the right direction. I'll start on that after this pull request goes through if that's okay.

user_sync/app.py
ianmak: ah yeah I added it because as far as I could tell, the --generate-remove-list would only include federated users... --generate-delete-list also includes orphaned managed users. Should I remove this?

user_sync/rules.py
```diff
@@ -62,11 +68,26 @@ def __init__(self, caller_options):
...
+        delete_user_key_list = set(delete_user_key_list) if (delete_user_key_list != None) else set()
+        self.delete_user_key_list = delete_user_key_list
+        '''
+        
+        # we only need a reference to either the remove or delete key list,
+        # since you can't do both
+        remove_or_delete_user_key_list = options['remove_user_key_list'] if options['remove_user_key_list'] != None else options['delete_user_key_list']
+        remove_or_delete_user_key_list = set(remove_or_delete_user_key_list) if (remove_or_delete_user_key_list != None) else set()
+        self.remove_or_delete_user_key_list = remove_or_delete_user_key_list
+
```
adobeDan: add a variable:
```py
orphan_list_output_path = options['remove_list_output_path] or ['delete_list_output_path']
self.orphan_list_output_path = orphan_list_output_path
```

user_sync/rules.py
```diff
@@ -291,24 +312,35 @@ def is_selected_user_key(self, user_key):
         return True
     
     def process_orphaned_dashboard_users(self):
-        remove_user_key_list = self.remove_user_key_list
+        remove_or_delete_user_key_list = self.remove_or_delete_user_key_list
             
         options = self.options
         remove_list_output_path = options['remove_list_output_path']
```
adobeDan: change all this to use `orphan_list_output_path`

user_sync/rules.py
```diff
@@ -291,24 +312,35 @@ def is_selected_user_key(self, user_key):
...
+        elif self.need_to_process_orphaned_dashboard_users_for_delete:
+            orphaned_dashboard_users = list(self.iter_orphaned_dashboard_users(self.options['all_identity_types']))
+            self.logger.info('Orphaned users to be deleted: %s', [self.get_dashboard_user_key(dashboard_user) for dashboard_user in orphaned_dashboard_users])
+        else:
+            raise user_sync.error.AssertionException('User removal operation type invalid.')            
         
         number_of_orphaned_dashboard_users = len(orphaned_dashboard_users)
 
-        if (remove_list_output_path != None):
+        if remove_list_output_path != None:
```
adobeDan: Great that you removed the parens, but...
- You should never say != None, the correct Python "is not None"
- You shouldn't ever test for is not None when you are really asking "does it have a value", which is what you are asking in this case, you should just say:
```py
if remove_list_output_path:
```

adobeDan: Great that you removed the parens, but...
- You should never say != None, the correct Python is "is not None"
- You shouldn't ever test for is not None when you are really asking "does it have a value", which is what you are asking in this case, you should just say:
```py
if remove_list_output_path:
```

user_sync/rules.py
```diff
@@ -291,24 +312,35 @@ def is_selected_user_key(self, user_key):
...
+        if self.need_to_process_orphaned_dashboard_users_for_remove:
+            orphaned_dashboard_users = list(self.iter_orphaned_dashboard_users(self.options['managed_identity_types']))
+            self.logger.info('Orphaned users to be removed: %s', [self.get_dashboard_user_key(dashboard_user) for dashboard_user in orphaned_dashboard_users])
+        elif self.need_to_process_orphaned_dashboard_users_for_delete:
+            orphaned_dashboard_users = list(self.iter_orphaned_dashboard_users(self.options['all_identity_types']))
+            self.logger.info('Orphaned users to be deleted: %s', [self.get_dashboard_user_key(dashboard_user) for dashboard_user in orphaned_dashboard_users])
+        else:
+            raise user_sync.error.AssertionException('User removal operation type invalid.')            
         
         number_of_orphaned_dashboard_users = len(orphaned_dashboard_users)
```
adobeDan: collapse the next two ifs to use `orphan_list_output_path` and generic language about "Writing orphaned users to ..."

user_sync/rules.py
```diff
@@ -347,9 +379,10 @@ def try_and_remove_from_org(user_key):
                 if (not owning_organization_info.is_dashboard_users_loaded() or owning_organization_info.get_dashboard_user(user_key) != None):
                     self.logger.info('Removing user for user key: %s', user_key)
                     id_type, username, domain = self.parse_user_key(user_key)
+                    do_delete_user = self.delete_after_remove_from_org and id_type != user_sync.identity_type.ADOBEID_IDENTITY_TYPE
```
adobeDan: The `and` is not needed here, `remove_from_org` will do the right thing on Adobe IDs without a complaint, even if you say `do_delete_user=True`.

user_sync/rules.py
```diff
@@ -780,14 +812,14 @@ def read_remove_list(file_path, delimiter = None, logger = None):
             id_type = row.get(id_type_column_name)
             user = row.get(user_column_name)
             domain = row.get(domain_column_name)
-            user_key = RuleProcessor.get_user_key(user, domain, id_type)
+            user_key = RuleProcessor.get_user_key(user, domain, None, id_type)
```
phil-levy: Could use some comments here.  What is this doing?

adobeDan: This can't be right.  get_user_key doesn't take four arguments.  What?

user_sync/rules.py
```diff
@@ -798,8 +830,11 @@ def write_remove_list(self, file_path, dashboard_users):
                 id_type, username, domain = self.parse_user_key(user_key)
                 writer.writerow({'type': id_type, 'user': username, 'domain': domain})
                 total_users += 1
-        self.logger.info('Total users in remove list: %d', total_users)
-
+        if is_delete_list:
```
adobeDan: See comment about about using generic language.

user_sync/rules.py
```diff
@@ -780,14 +812,14 @@ def read_remove_list(file_path, delimiter = None, logger = None):
...
-            user_key = RuleProcessor.get_user_key(user, domain, id_type)
+            user_key = RuleProcessor.get_user_key(user, domain, None, id_type)
             if user_key:
                 result.append(user_key)
             elif logger:
                 logger.error("Invalid input line, ignored: %s", row)
         return result
 
-    def write_remove_list(self, file_path, dashboard_users):
+    def write_remove_list(self, file_path, dashboard_users, is_delete_list=False):
```
adobeDan: Don't need this if we use generic language.

user_sync/rules.py
ianmak: I added the None parameter because get_user_key requires 4 parameters, and read_remove_list was crashing when it got to this point... the email parameter for get_user_key doesn't seem to be required, so I set the argument to None for now...

ianmak: I thought it didn't, but in the current v1 it does seem to (it now wants an email input as well?). It was crashing in read_remove_list, because it complained it only provided 3 arguments but expected 4.

adobeDan: Right, I get it.  The bug is that get_user_key doesn't specify domain and email as optional.  You should actually fix that (on line 721) and then call get_user_key with keyword args here.  But just leave it for now and I'll fix it everywhere as a separate change.

Be sure to put in a comment saying "specifying None for (optional) email, because we don't have it".

README.md
```diff
@@ -96,26 +96,23 @@ optional arguments:
...
                         in mapped groups matches the customer side.
   --remove-nonexistent-users
-                        Causes the user sync tool to remove Federated users
-                        that exist on the Adobe side if they are not in the
-                        customer side AD. This has the effect of deleting the
-                        user account if that account is owned by the
-                        organization under which the sync operation is being
-                        run.
+                        Causes the user sync tool to remove users that exist on
+                        the Adobe side if they are not in the customer side AD.
```
adobeDan: customer side directory.  Try and Find All on "AD" (case sensitive) and remove everyone you find! :)

README.md
```diff
@@ -125,8 +122,8 @@ optional arguments:
   --delete-list input_path
                         specifies the file containing the list of users to be
                         deleted. Enterprise and federated users on this list
-			are deleted on the Adobe side, and Adobe ID's are
-			removeFromOrg'd.
+                        are deleted on the Adobe side, and Adobe ID's are
```
adobeDan: Adobe ID user accounts, not "Adobe ID's".

README.md
```diff
@@ -125,8 +122,8 @@ optional arguments:
   --delete-list input_path
                         specifies the file containing the list of users to be
```
adobeDan: user accounts

docs/user-manual/index.md
```diff
@@ -787,12 +787,12 @@ specific behavior in various situations.
...
-| `--generate-delete-list` _output\_path_ | When supplied, all users found on the Adobe side that are not in the enterprise directory, lists those users to the given file. You can then pass this file to the `delete-list` argument in a subsequent run. |
-| `--delete-list` _input\_path_ | Deletes a list of users contained in the given file from the Adobe organization. Adobe ID users are removed from the organization. |
+| `--remove-nonexistent-users` | When supplied, if Enterprise ID or Federated ID user accounts are found on the Adobe side that are not in the customer-side directory, removes those user accounts from the organization. |
+| `--generate-remove-list` _output\_path_ | When supplied, if Enterprise ID or Federated ID user accounts are found on the Adobe side that are not in the enterprise directory, lists those users to the given file. You can then pass this file to the `remove-list` argument in a subsequent run. |
+| `--remove-list` _input\_path_ | Removes a list of users contained in the given file from the Adobe-side directory. |
```
adobeDan: Adobe-side organization (not directory)

README.md
```diff
@@ -112,6 +112,21 @@ optional arguments:
...
```
ianmak: Oops let the "AD" slip by. Corrected the occurrence in the next commit.

docs/user-manual/index.md
```diff
@@ -968,7 +968,7 @@ is provided with the tool.
 #### Add users and generate a list of users to remove
 
 This action synchronizes all users from the customer side with
-the Adobe side and also generates a list of federated users that no longer
+the Adobe side and also generates a list of users that no longer
 exist in the enterprise directory but still exist on the Adobe
```
adobeDan: customer directory, not enterprise directory

user_sync/rules.py
```diff
@@ -68,27 +65,21 @@ def __init__(self, caller_options):
...
-        self.remove_or_delete_user_key_list = remove_or_delete_user_key_list
+        self.orphan_user_key_list = set(options['remove_user_key_list'] or options['delete_user_key_list'] or [])
+        self.orphan_list_output_path = options['remove_list_output_path'] or options['delete_list_output_path']
+        
+        # determine if we need to delete user accounts instead of just removing
+        # them
+        self.delete_user_accounts = options['delete_nonexistent_users'] or options['delete_user_key_list']
+        
+        # determine how we should treat orphaned accounts during processing
+        self.process_orphaned_accounts_for_delete = options['delete_list_output_path'] is not None or options['delete_nonexistent_users']
```
adobeDan: take out "is not None", it's not necessary (and is overly strict, even if delete_list_output_path was specified as an empty string we want the test to fail).  Thus = "options[...] or options[...]" not "options[...] is not None or options[...]".

user_sync/rules.py
```diff
@@ -68,27 +65,21 @@ def __init__(self, caller_options):
...
+        
+        # determine whether we need to process orphaned users at all
+        self.need_to_process_orphaned_dashboard_users = options['remove_list_output_path'] is not None or options['remove_nonexistent_users'] or self.process_orphaned_accounts_for_delete
```
adobeDan: remove "is not None".  see above.

README.md
```diff
@@ -96,22 +96,35 @@ optional arguments:
...
+  --generate-delete-list output_path
+                        processing similar to --delete-nonexistent-users
+                        except that rather than performing removals, a file is
+                        generated (with the given pathname) listing users who
+                        would be removed. This file can then be given in the
+                        --delete-list argument in a subsequent run.
+  --delete-list input_path
+                        specifies the file containing the list of users to be
+                        deleted. Enterprise and federated users on this list
+                        are deleted on the Adobe side, and Adobe ID's are
```
adobeDan: Adobe ID users, not "Adobe ID's".

docs/success-guide/command_line_options.md
```diff
@@ -45,7 +45,8 @@ Note that only the users returned by the directory query and filter are consider
 |   `--disable-nonexistent-users`\* |    Adobe account to remain but licenses and group <br>memberships are removed.  |
 |   `--remove-nonexistent-users`  |    Adobe account to remain but licenses, group memberships, and membership in this org to be removed   |
 |   `--delete-nonexistent-users`\*  |    Adobe account to be deleted: remove from PLCs and user groups and <br>from the org; account deleted and all storage and settings freed. |
-|   `--generate-remove-list f`    |  No action to be taken on the account.  User name written to file for later action. |
+|   `--generate-remove-list f`    |  No action to be taken on the account.  Federated user name written to file for later action. |
```
adobeDan: No "Federated".  Just "User name written"

docs/user-manual/index.md
```diff
@@ -944,24 +947,61 @@ such a file, example.users-file.csv, is provided with the tool.
...
 file generated by the generate-remove-list parameter, or you
 can pass a CSV file of users that you have generated by some
 other means An example of such a file, example.users-file.csv,
 is provided with the tool.
 
+#### Add users and generate a list of users to remove
+
+This action synchronizes all users from the customer side with
+the Adobe side and also generates a list of users that no longer
+exist in the enterprise directory but still exist on the Adobe
```
adobeDan: customer directory, not enterprise directory

docs/user-manual/index.md
```diff
@@ -944,24 +947,61 @@ such a file, example.users-file.csv, is provided with the tool.
...
+#### Remove users from separate list
+
+This action takes a CSV file containing a list of users that have
+been flagged for removal, and removes those users from the organization
+on the Adobe side. The CSV file is typically the one generated by a
+previous call that used the generate-remove-list parameter.
+
+You can create a CSV file of users to remove by some other means.
+However, if your list includes any users that still exist in your
+enterprise directory, those users will be added back in on the
```
adobeDan: customer directory, not enterprise directory

user_sync/app.py
```diff
@@ -72,9 +72,18 @@ def process_args():
     parser.add_argument('--generate-remove-list',
                         help='processing similar to --remove-nonexistent-users except that rather than performing removals, a file is generated (with the given pathname) listing users who would be removed. This file can then be given in the --remove-list argument in a subsequent run.',
                         metavar='output_path', dest='remove_list_output_path')
-    parser.add_argument('-d', '--remove-list',
-                        help='specifies the file containing the list of users to be removed. Users on this list are removeFromOrg\'ed on the Adobe side.',
+    parser.add_argument('--remove-list',
+                        help='specifies the file containing the list of users to be removed. Users on this list are removed from the organization on the Adobe side.',
                         metavar='input_path', dest='remove_list_input_path')
+    parser.add_argument('--delete-nonexistent-users',
+                        help='Causes the user sync tool to delete users that exist on the Adobe side if they are not in the customer side AD. If the account is an Adobe account, it is remove from org\'d, but not deleted.',
```
adobeDan: users -> user accounts,
Adobe account -> Adobe ID account
remove from org'd -> removed from the Adobe-side organization

user_sync/app.py
```diff
@@ -72,9 +72,18 @@ def process_args():
...
+                        action='store_true', dest='delete_nonexistent_users')
+    parser.add_argument('--generate-delete-list',
+                        help='processing similar to --delete-nonexistent-users except that rather than performing deletions, a file is generated (with the given pathname) listing users who would be deleted. This file can then be given in the --delete-list argument in a subsequent run.',
+                        metavar='output_path', dest='delete_list_output_path')
+    parser.add_argument('--delete-list',
+                        help='specifies the file containing the list of users to be deleted. Users in this list are removed from the organization on the Adobe side, and managed users are deleted on the Adobe side.',
```
adobeDan: managed users -> Enterprise and Federated

user_sync/rules.py
```diff
@@ -269,12 +281,12 @@ def process_dashboard_users(self, dashboard_connectors):
                 for user_key, desired_groups in accessor_unprocessed_groups_by_user_key.iteritems():
                     self.try_and_update_dashboard_user(accessor_organization_info, user_key, dashboard_connector, groups_to_add=desired_groups)
                     
-    def iter_orphaned_dashboard_users(self, orphan_account_types):
+    def iter_orphaned_dashboard_users(self, orphan_account_types=None):
```
adobeDan: Don't add a default, we don't need it.  and the one that you added makes this function a no-op, so there's no point in calling it.

user_sync/rules.py
```diff
@@ -269,12 +281,12 @@ def process_dashboard_users(self, dashboard_connectors):
...
         owning_organization_info = self.get_organization_info(OWNING_ORGANIZATION_NAME)
         for user_key, dashboard_user in owning_organization_info.iter_orphaned_dashboard_users():
             if not self.is_selected_user_key(user_key):
                 continue
-            if (dashboard_user.get('type') not in orphan_account_types):
+            if orphan_account_types is not None and dashboard_user.get('type') not in orphan_account_types:
```
adobeDan: take out this test you added.  the caller must always provide a set or list.

user_sync/rules.py
```diff
@@ -291,24 +303,29 @@ def is_selected_user_key(self, user_key):
...
         remove_nonexistent_users = options['remove_nonexistent_users']
+        delete_nonexistent_users = options['delete_nonexistent_users']
         
         max_deletions_per_run = options['max_deletions_per_run']
         max_missing_users = options['max_missing_users']
 
-        orphaned_dashboard_users = list(self.iter_orphaned_dashboard_users(self.options['managed_identity_types']))
-        self.logger.info('Orphaned users to be removed: %s', [self.get_dashboard_user_key(dashboard_user) for dashboard_user in orphaned_dashboard_users])
+        if self.process_orphaned_accounts_for_delete:
+            orphaned_dashboard_users = list(self.iter_orphaned_dashboard_users())
```
adobeDan: this should be `self.iter_orphaned_dashboard_users(self.options['managed_identity_types'])`

I think maybe you don't understand the `managed_identity_types` option.

ianmak: Oh darn I just realized, the Adobe ID's would never show up as an orphaned account... is that why it's only looking for managed accounts? So the only way to delete Adobe ID accounts would be to specifically list them?

adobeDan:
@ianmak Don't forget to pull your merged PR from v1 back into your branch; it's required in order for the PR to be merged.  I recommend you do a `git pull --rebase origin v1` so it doesn't actually show the merge as an extra commit.

user_sync/rules.py
adobeDan: No, that's not it.  The `managed_identity_types` option in the config file, which defaults to `[federatedID, enterpriseID]`, is what controls which kinds of accounts can be removed/deleted.  So if you alter your config to add `adobeID` to that list then this code will also delete the Adobe IDs.  That's why we make the back end smart about just doing removeFromOrg on Adobe IDs if the client asks for them to be deleted.

ianmak: Ohh okay I get it now. I thought remove and delete were filtering accounts differently for some weird reason. I'll update to correct this.

ianmak:
oh shoot, I did the `git pull --rebase origin v1`, but then on github desktop accidentally clicked "update from v1", now the merge from v1 shows up on the branch... would you happen to know how to undo that?

adobeDan:
Don’t sweat it.  It’s an empty commit; no harm it being there :). –d.

Issue #49, opened 2017-03-03
Unhandled exception when slight error in config file


phil-levy:
If you have
limits:
    max_deletions_per_run= 10    

you get
2017-03-02 17:31:55 17952 ERROR main - Unhandled exception
Traceback (most recent call last):
  File "c:\user-sync\.pex\install\user_sync-1.1.1-py2-none-any.whl.1d38cd7281dcf25d7c48135e8d87ce573a9f8c9f\user_sync-1.1.1-py2-none-any.whl\user_sync\app.py", line 256, in main
    config_loader = create_config_loader(args)
  File "c:\user-sync\.pex\install\user_sync-1.1.1-py2-none-any.whl.1d38cd7281dcf25d7c48135e8d87ce573a9f8c9f\user_sync-1.1.1-py2-none-any.whl\user_sync\app.py", line 184, in create_config_loader
    config_loader = user_sync.config.ConfigLoader(config_bootstrap_options)
  File "c:\user-sync\.pex\install\user_sync-1.1.1-py2-none-any.whl.1d38cd7281dcf25d7c48135e8d87ce573a9f8c9f\user_sync-1.1.1-py2-none-any.whl\user_sync\config.py", line 75, in __init__
    config_content = self.load_from_yaml(self.main_config_path)
  File "c:\user-sync\.pex\install\user_sync-1.1.1-py2-none-any.whl.1d38cd7281dcf25d7c48135e8d87ce573a9f8c9f\user_sync-1.1.1-py2-none-any.whl\user_sync\config.py", line 260, in load_from_yaml
    return yaml.load(input_file)
  File "c:\user-sync\.pex\install\pyyaml-3.12-cp27-cp27m-win_amd64.whl.95f7f7154d83db289311a6943ed874bc14dae938\pyyaml-3.12-cp27-cp27m-win_amd64.whl\yaml\__init__.py", line 71, in load
    return loader.get_single_data()
  File "c:\user-sync\.pex\install\pyyaml-3.12-cp27-cp27m-win_amd64.whl.95f7f7154d83db289311a6943ed874bc14dae938\pyyaml-3.12-cp27-cp27m-win_amd64.whl\yaml\constructor.py", line 37, in get_single_data
    node = self.get_single_node()
[...]

PR #82, opened 2017-03-15
fix for unhandled exception when slight error in config file

ianmak:
config parser errors now have the following form:

```log
2017-03-15 14:10:50 47278 CRITICAL main - Error loading configuration
file: while parsing a block mapping
in "user-sync-config.yml", line 1, column 1
expected <block end>, but found '<block mapping start>'
in "user-sync-config.yml", line 104, column 5
```

### DIFF
--- a/user_sync/config.py
+++ b/user_sync/config.py
@@ -257,11 +257,15 @@ class ConfigLoader(object):
         try:
             with open(file_path, 'r', 1) as input_file:
                 return yaml.load(input_file)
+        except IOError as e:
+            # if a file operation error occurred while loading the
+            # configuration file, swallow up the exception and re-raise this
+            # as an configuration loader exception.
+            raise user_sync.error.AssertionException('Error reading configuration file: %s' % e)
         except yaml.error.MarkedYAMLError as e:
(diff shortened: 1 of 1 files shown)


user_sync/config.py
```diff
@@ -254,8 +254,14 @@ def load_from_yaml(self, file_path):
...
-        with open(file_path, 'r', 1) as input_file:
-            return yaml.load(input_file)
+        try:
+            with open(file_path, 'r', 1) as input_file:
+                return yaml.load(input_file)
+        except yaml.error.MarkedYAMLError as e:
+            # if a parser error occured while loading the configuration file,
+            # swallow the error and re-raise it as a configuration loader
+            # exception
+            raise user_sync.error.AssertionException('Error loading configuration file: %s' % e)
```
adobeDan: Please make clear that it was a YAML syntax error, not some sort of general error.
Also please catch any file-open exceptions unless there's some other code that already ensures the file is readable.

phil-levy: What Dan said + many thanks for fixing this!

adobeDan:
Fixes #49.

Issue #83, opened 2017-03-16
documentation update


Luci2015:
Please update the pdf and word documentation on page 10
in "Example configurations, under user-sync.config.yml:
 “dashboard" should not contain attribute “user-identity-type”. Should be under "directory"

Thanks!

adobeDan:
I believe we should patch the published docs with this change in advance of the next release.  That will mean changing out the PDF to be one created from the updated doc set.  @phil-levy we should discuss how we want to create the PDF version.

Issue #76, opened 2017-03-14
Document user_username_format and user_domain_format


adorton-adobe:
I don't see any mention of user_username_format or user_domain_format in the Sync Guide.  I'm helping a customer that uses username-based auth on their federated domain, and we had to search the source code to figure out how to set username and domain on user records.

adobeDan:
Yep, great point.  See also #48: I'm doing work there to add a `user_identity_type_format` so you can have different types of users in your one directory system.  We will need to document this with those others when added.

In general our docs are not very good about how to work with username-based federation.  And there are a lot of bugs about that in the current release, all of which I think are fixed in the v1 branch at this point.

adobeDan:
@phil-levy Can we have @jabogart take a crack at this?

phil-levy:
What is the the problem?  Also why do we want more than one identity type in a directory.  Seems like that would be very problematic for users.  Is this really a multi-domain thing?

adobeDan:
@phil-levy I think you meant your comment for #48.  Happy to discuss there.  For this issue, it's just one of updating the docs to explain how to set up your connector-ldap for the cases where you are doing federated with username-based login.  You have to have attributes for the username and the domain, as well as the email.

phil-levy:
Who knows the answer about  non-email username?  We'd have to gather the information.

adobeDan:
Both Andrew and I know the answer.  You have to set the value of `user_username_format` and (optionally)`user_domain_format` in your connector-ldap.yaml config file.  You set them to a string of the form `'{attrname}'` where the value of the `attrname` attribute is the value to use, or if it's a fixed value (as it often is for domains), then you set the format string to be that value, as in `'domain.com'`.  When reading the directory, the connector will notice that the username field does or does not have an `@` in it, and if it does then it will assume that it's an email-login federated user and take the domain from the email.  if there's no `@` in the username, then it will use the domain value as the domain (so in that case it's not optional that you specify it).  Does that help?

adobeDan:
@phil-levy Can you get to this before you leave on vacation or do you need me to do the doc update?

phil-levy:
I can do.

adobeDan:
Fixed in PR #86 but I screwed up the commit message so it didn't auto-close.

PR #86, opened 2017-03-16
Doc update for username based login and fix config example

phil-levy:


phil-levy:
Fix #83 
Fix #76

docs/user-manual/index.md
```diff
@@ -1356,6 +1369,28 @@ group in its group map.  It updates membership in the user group,
 which indirectly updates the membership in the product
 configuration.
 
+### Working with Non-Email-based Usernames
+
+On the Adobe Admin Console, you can configure a domain to use email-based user login names or non-email-based usernames.  We often user the term Username based login to mean non-email-based-login.  Username-based login can be used when email addresses are expected to change often or your organization does not allow email addresses to be used for login.
```
adobeDan: This needs some work.  See if you can work in the following information:
* This only applies to Federated ID domains, that use SAML-based SSO with an IdP.
* The terms Username-based login and Email-based login come directly from the SSO configuration settings for the domain, where there is a configuration field called "User Login Setting" that can be set to "Email address" or "Username" (see snapshot below)
* Whether to use username-based login or email-based login depends on a company's overall identification strategy, and are outside the scope of this document.

sso federation snapshot

docs/user-manual/index.md
```diff
@@ -1356,6 +1369,28 @@ group in its group map.  It updates membership in the user group,
...
+In the `connector-ldap.yml` file:
+
+- Set the value of `user_username_format` to a value like '{attrname}' where attrname names the directory attribute whose value is to be used for the user name.
+- Set the value of `user_domain_format` to a value like '{attrname}' if the domain name comes from the named directory attribute, or to a fixed string value like 'example.com'.
+
+When processing the directory, User Sync will fill in the username and domain values from those fields (or values).
+
+The values given for these configuration items can be a mix of string characters and one or more attribute names enclosed in curly-braces "{}".  The fixed characters are combined with the attribute value to form the string used in processing the user.
+
+Note that `user_username_format` configuration item will need to produce an email address if the federated domain is using email-based login. That can happen because the field itself contains an email address, or because you set the format string to something like `{attrname}@example.com`.
```
adobeDan: I think you should remove this paragraph completely, because the whole premise of the section is "how to set up for username-based login", and this paragraph is about "how to set up for email-based login :).  If you want to talk about email-based login in this section, I would just add a paragraph at the very bottom that says:

If your Federated ID domain uses email-based login, then you should not define either of the parameters describe here: User Sync will automatically use the user's email value for SSO purposes.

docs/user-manual/index.md
```diff
@@ -1356,6 +1369,28 @@ group in its group map.  It updates membership in the user group,
...
+
+For domains that use username-based login, the `user_username_format` configuration item should not produce an email address, and that happen because the attribute you choose likely doesn't contain one.  Note that the "@" character is not allowed in usernames used in username-based login.
+
+
```
adobeDan: Add a paragraph: Keep in mind that, even if you are using username-based login, you *must* provide an email address for every user, and that email address must be unique for every user.  User Sync will never add a user to the Adobe side without an email address.

Issue #45, opened 2017-03-02
Add version and invocation parameters to the log


phil-levy:
It is hard to diagnose problem when customers provide logs without this.

PR #91, opened 2017-03-17
Issue #45: add version number and invocation parameters to log

ensembledeanw:


user_sync/app.py
```diff
@@ -299,6 +299,17 @@ def create_config_loader_options(args):
     
     return config_options
 
+def log_parameters(args):
+    '''
+    Log the invocation parameters to make it easier to diagnose problem when customers
```
adobeDan: Looks like you lost some text here. "to diagnose problem when customers"?

user_sync/app.py
```diff
@@ -299,6 +299,17 @@ def create_config_loader_options(args):
...
+    :param args: namespace
+    :return: None
+    '''
+    logger.info('----- Invocation Parameters -----')
+    for parameter_name, parameter_value in args.__dict__.iteritems():
+        logger.info('  %s: %s', parameter_name, parameter_value)
+    logger.info('---------------------------------')
```
adobeDan: So this looks like it logs our internal (code-level) parameter names and values, which is good, but it doesn't get at the essence of the ask from a customer perspective, which is to actually put the command line that the customer gave us in the log.  Can you please *add* to what you have here additional code which essentially logs the sys.argv command line?

user_sync/app.py
```diff
@@ -309,9 +320,11 @@ def main():
         
         config_loader = create_config_loader(args)
         init_log(config_loader.get_logging_config())
-        
-        run_stats = user_sync.helper.JobStats("Run", divider = "=")
+
+        # add start divider, app version number, and invocation parameters to log
+        run_stats = user_sync.helper.JobStats('Run (Ver: ' + APP_VERSION + ')', divider='=')
```
adobeDan: How about "User Sync version:"  No need to be terse, and much clearer for non-English speakers then "Ver".

ensembledeanw:
Thank you for your comments, Dan. I made the changes base on your suggestion and it is ready for review again.

PR #96, opened 2017-03-19
Fix #87.  Adds `exclude_groups` and `exclude_users`.

### LINKED ISSUE #87: User protection feature
Create a feature to protect Adobe users from being updated or removed by User Sync.  The purpose of this feature is to provide a means for customers to prevent User Sync from interfering with users that they are administering manually.

We extend the existing `dashboard` configuration with the ability to identify Adobe accounts that are immune from update, group change, and removal.  Users so identified are protected from any alteration by User Sync including
     --remove-nonexistent-users, 
     --delete-nonexistent-users,  
     --remove-entitlements-for-nonexistent-users 
     --update-user-info - updating fields of the user
     --process-groups  - updating group membership
     --remove-list (etc)

The additional configuration options under dashboard are as follows:
```YAML
dashboard:
  owning: dashboard-config.yml  # existing key
  exclude_groups:  # new key: list of dashboard-side groups (user group or product configuration)
    - special_users   # members of any group in this list are excluded from updates/remove/etc.
    - more special users   # can have more than one group in a list
  exclude_users:   # new key: list of regular expressions matched against dashboard-side usernames
    - ".*@example.com"  # any matching user is excluded from updates/remove/etc.
    - pattern2                   # can have more than one regular expression in a list
  exclude_identity_types: # new key: list of identity types
    - adobeID                     # default in config file: means exclude Adobe IDs from update/remove/etc. 
    - enterpriseID               # can exclude additional identity types
    - federatedID
```
`exclude_groups` defines a list of Adobe user groups or PCs.  Adobe users who are members of such groups are not removed, updated, nor have their group membership changed.  

`exclude_users` gives a list of patterns.  Adobe users with usernames that match (default case-insensitive, unless the pattern specifies case-sensitive) any of the specified patterns (if there are more than one) are not removed, updated, nor have their group membership changed.

`exclude_identity_types` gives a list of identity types.  Adobe users who have one of these identity types are not removed, updated, nor have their group membership changed.

Note that directory side users are still created on the Adobe side even if one of the `exclude-` parameters would then exclude the user on the Adobe side from updates in successive runs.  That is, these parameters apply only to Adobe users who exist *when the Adobe directory is being matched against the customer directory*.

All of these additional configuration options are optional and default to empty (that is, no dashboard users are excluded from update/remove/etc.).  They should appear in the example main config file commented out, except for `exclude_identity_types`, which should have the sole member value `adobeID`.

Accounts that would have been removed or updated but were not because of this feature should be listed as info level log entries.

adobeDan:
* check for and prevent excluding based on non-owning org groups
* require the regular expressions to match the entire name
* use Unicode rules in regular expression matches.


user_sync/config.py
```diff
@@ -331,15 +327,36 @@ def get_rule_options(self):
...
         for name in exclude_identity_type_names:
             identity_type = user_sync.identity_type.parse_identity_type(name)
             if not identity_type:
-                validation_message = "Illegal value for %s in config file: %s" % ('exclude_identity_types', name)
+                validation_message = 'Illegal value for %s in config file: %s' % ('exclude_identity_types', name)
                 raise user_sync.error.AssertionException(validation_message)
             exclude_identity_types.append(identity_type)
+        for regexp in exclude_users_regexps:
+            try:
+                exclude_users.append(re.compile(r'\A' + regexp + r'\Z', re.UNICODE))
```
phil-levy: Could use a comment about what this is testing for

adobeDan:
I realized this afternoon that there are bugs in the current logic when you get past the owning org.  I'm going to check in a fix for them in this changelist before I merge it.  @phil-levy if you get a chance to review, fine, otherwise I'll review in my ancientHacker alter ego :).

PR #99, opened 2017-03-21
Issue 30/file paths in config files should be interpreted relative to the file in which they appear 2

ianmak:
* Add support for an identity_type (customizable) attribute definition a la email and others.
* Regularize spelling to 'identity_type' everywhere in directory/CSV user dictionaries.

### DIFF
--- a/user_sync/config.py
+++ b/user_sync/config.py
@@ -30,7 +30,6 @@ import user_sync.error
 import user_sync.identity_type
 import user_sync.rules
 
-DEFAULT_CONFIG_DIRECTORY = ''
 DEFAULT_MAIN_CONFIG_FILENAME = 'user-sync-config.yml'
 DEFAULT_DASHBOARD_OWNING_CONFIG_FILENAME = 'dashboard-owning-config.yml'
 DEFAULT_DASHBOARD_ACCESSOR_CONFIG_FILENAME_FORMAT = 'dashboard-accessor-{organization_name}-config.yml'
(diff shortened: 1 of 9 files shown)


user_sync/config.py
```diff
@@ -129,7 +126,7 @@ def get_dashboard_options_for_accessors(self):
             accessor_config = None
             if (accessors_config != None): 
                 accessor_config = accessors_config.get_list(organization_name, True) 
-            accessor_config_sources = self.as_list(accessor_config)
+            accessor_config_sources = ConfigFileLoader.get_relative_filenames(self.as_list(accessor_config))
```
adobeDan: I don't think this is right; it presumes that all of the configs are filenames, but of course they may not always be filenames.  This should move down, see below.

user_sync/config.py
```diff
@@ -233,7 +230,7 @@ def get_dict_from_sources(self, sources, owner):
             if (isinstance(source, types.StringTypes)):
                 absolute_path = self.get_absolute_file_path(source)
```
adobeDan: This is where you should be calling `ConfigFileLoader.get_relative_filename` on the `source`, rather than the `get_absolute_file_path` that's being called.  But it might be easier just to change self.get_absolute_file_path to do the call to ConfigFileLoader.

user_sync/config.py
ianmak: oh darn good call, I'll correct this

adobeDan:
Can you call me 408-536-4150?  I want to check in about status. –d.

Issue #30, opened 2017-02-28
file paths in config files should be interpreted relative to the file in which they appear


phil-levy:
If there is an entry in the dashboard-config.yml file like this:

   priv_key_path: private.key

then the file should be interpreted as relative to the directory containing the dashboard-config.yml file.  Now it appears it is relative to the current directory so that if there is a -c option for the config file location, this file ends up not being found unless it is fully qualified.

phil-levy:
I believe this error also occurs for other file references.  All unqualified file references in the configuration files should be relative to the file containing them.  Otherwise, you can't really use the -c option to point to a set of files outside of the current directory.

adobeDan:
@ianmak A more limited version of this is to take all pathnames relative to the location of the main configuration file.  Let's go with this version first, because it's a simpler change for 1.2rc1, and then we will look at making the full change for 1.2 GM.

PR #102, opened 2017-03-22
added support for yaml file path processing

ianmak:
additional support for issue #30

### DIFF
--- a/user_sync/config.py
+++ b/user_sync/config.py
@@ -39,17 +39,17 @@ DEFAULT_DASHBOARD_ACCESSOR_CONFIG_FILENAME_FORMAT = 'dashboard-accessor-{organiz
 ROOT_CONFIG_PATH_KEYS = [
         { 'path':'/dashboard/owning', 'default': DEFAULT_DASHBOARD_OWNING_CONFIG_FILENAME },
         '/dashboard/owning/enterprise/priv_key_path',
-        '/dashboard/accessors',
+        '/dashboard/accessors/*',
         '/dashboard/accessors/*/enterprise/priv_key_path',
         { 'path':'/dashboard/accessor_config_filename_format', 'default':DEFAULT_DASHBOARD_ACCESSOR_CONFIG_FILENAME_FORMAT },
-        '/directory/connectors/ldap'
+        '/directory/connectors/ldap',
+        '/logging/file_log_directory'
     ]
 
 # like ROOT_CONFIG_PATH_KEYS, but this is applied to non-root configuration
 # files
 SUB_CONFIG_PATH_KEYS = [
-        "/enterprise/priv_key_path",
-        "/logging/file_log_directory"
+        "/enterprise/priv_key_path"
     ]
 
 class ConfigLoader(object):
@@ -623,8 +623,8 @@ class ConfigFileLoader(object):
         :rtype dict
         '''
         try:
-            input_file = open(filename, 'r', 1)
-            yml = yaml.load(input_file)
+            with open(filename, 'r', 1) as input_file:
+                yml = yaml.load(input_file)
         except IOError as e:
             # if a file operation error occurred while loading the
             # configuration file, swallow up the exception and re-raise this
@@ -645,6 +645,9 @@ class ConfigFileLoader(object):
             to begin with. It is used recursively to navigate into child
             dictionaries to search the path key.
             type dictionary: dict
+            type keys: list
+            type level: int
+            type default_val: str
             '''
             def relative_path(filename):
                 '''
@@ -670,11 +673,13 @@ class ConfigFileLoader(object):
                     if isinstance(val,str):
                         dictionary[key] = relative_path(val)
                     elif isinstance(val,list):
-                        paths = []
-                        for path in val:
-                            if isinstance(path, str):
-                                paths.append(relative_path(path))
-                        dictionary[key] = paths
+                        vals = []
+                        for entry in val:
+                            if isinstance(entry, str):
+                                vals.append(relative_path(entry))
+                            else:
+                                vals.append(entry)
+                        dictionary[key] = vals
 
             # end of path key
             if level == len(keys)-1:
--- a/user_sync/config.py
+++ b/user_sync/config.py
@@ -34,16 +34,22 @@ DEFAULT_MAIN_CONFIG_FILENAME = 'user-sync-config.yml'
 DEFAULT_DASHBOARD_OWNING_CONFIG_FILENAME = 'dashboard-owning-config.yml'
 DEFAULT_DASHBOARD_ACCESSOR_CONFIG_FILENAME_FORMAT = 'dashboard-accessor-{organization_name}-config.yml'
 
-# list of key paths that will be processed as filenames relative to the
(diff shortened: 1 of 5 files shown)


user_sync/config.py
```diff
@@ -246,17 +258,8 @@ def get_absolute_file_path(self, value):
         make sure the path is relative from the config path.
         :type value: str
         '''        
-        return ConfigFileLoader.get_relative_filename(value)
+        return os.path.abspath(value)
```
adobeDan: So the commentary on this function is wrong, and the function itself should go away.  You will only end up calling this once, once you have located the actual config file, so just call os.path.abspath directly at that point.

user_sync/config.py
```diff
@@ -83,7 +95,7 @@ def get_logging_config(self):
 
     def get_dashboard_options_for_owning(self):
         owning_config_filename = DEFAULT_DASHBOARD_OWNING_CONFIG_FILENAME
-        owning_config_path = self.get_file_path(owning_config_filename)
+        owning_config_path = self.get_absolute_file_path(owning_config_filename)
```
adobeDan: No, this has to be processed as if it were relative to the main config file.  All of the predefined default values have to be relativized as if they were part of the main config file.  At the time you find the main config file and process it with load-yaml, you will need to check to see if the user actually provided values for these constants in the config file.  If not, you will need to compute the right absolute pathnames for them (same directory as the config file but their value as the filename part) and then insert those keys into the result of load-yaml on the main config.  This is why I was recommending you leave the ConfigFileLoader class in place, because it still needs two entries: ConfigFileLoader.load_root_config which does the additional processing I'm describing here, and ConfigFileLoader.load_sub_config which only does what load_yaml does.  But if you want to get rid of ConfigFileLoader that's fine, you can just have two functions in the ConfigLoader class instead.

user_sync/config.py
```diff
@@ -316,43 +317,49 @@ def load_from_yaml(filename):
 
         dirpath = os.path.dirname(filename)
 
-        def process_dict_path_key(d, keys, level):
+        # this function is used to process a single path key by replacing the
```
adobeDan: In Python, function comments go as strings which are the first element in the function, and they also have annotations for the args and their types.  For example, see the comment on the function that contains this function.

This comment says what the function does, but a good function comment also says "how" the function does what it does.

user_sync/config.py
```diff
@@ -34,16 +34,22 @@
...
+        '/dashboard/accessors/*/enterprise/priv_key_path',
+        { 'path':'/dashboard/accessor_config_filename_format', 'default':DEFAULT_DASHBOARD_ACCESSOR_CONFIG_FILENAME_FORMAT },
+        '/directory/connectors/ldap'
+    ]
+
+# like ROOT_CONFIG_PATH_KEYS, but this is applied to non-root configuration
+# files
+SUB_CONFIG_PATH_KEYS = [
+        "/enterprise/priv_key_path",
+        "/logging/file_log_directory"
```
adobeDan: logging is actually in the ROOT_CONFIG. -d.

user_sync/config.py
```diff
@@ -694,6 +611,138 @@ def describe_unused_values(self):
...
+        '''
+        loads a yaml file, processes the resulting dict to adapt values for keys
+        (the path to which is defined in PATH_KEYS) to a value that represents
+        a file reference relative to provided source file, and returns the
+        processed dict.
+        :type filename: str
+        :rtype dict
+        '''
+        try:
+            input_file = open(filename, 'r', 1)
```
adobeDan: this doesn't do a close after the read.  the right way to do this is:
```python
with open(filename, 'r', 1) as input_file:
    yml = yaml.load(input_file)
```
That way the file is guaranteed to be closed on either success or error.

user_sync/config.py
```diff
@@ -694,6 +611,138 @@ def describe_unused_values(self):
...
+        dirpath = os.path.dirname(filename)
+    
+        def process_path_key(dictionary, keys, level, default_val):
+            '''
+            this function is used to process a single path key by replacing the
+            value for the key that is found into a path relative to the given
+            source file, with the assumption that the value is a file reference
+            to begin with. It is used recursively to navigate into child
+            dictionaries to search the path key.
+            type dictionary: dict
```
adobeDan: also have type references for key, level, default_val

user_sync/config.py
```diff
@@ -694,6 +611,138 @@ def describe_unused_values(self):
...
+                if dictionary.has_key(key):
+                    val = dictionary[key]
+                    if isinstance(val,str):
+                        dictionary[key] = relative_path(val)
+                    elif isinstance(val,list):
+                        paths = []
+                        for path in val:
+                            if isinstance(path, str):
+                                paths.append(relative_path(path))
+                        dictionary[key] = paths
```
adobeDan: Almost there!  But this is excluding all the dict values in `val`.  Since `val` is going to be processed by `get_dict_from_sources`, it can be a list of pathnames and dictionaries, and you have to preserve the dictionaries.  So it should be
```python
elif isinstance(val, list):
    vals = []
    for entry in val:
        if isinstance(entry, str):
            vals.append(relative_path(path))
        else:
            vals.append(entry)
    dictionary[key] = vals

File: user_sync/config.py [hunk 8]
```diff
@@ -34,16 +34,22 @@
...
-    "/dashboard/accessors/*/enterprise/priv_key_path",
-    "/dashboard/accessor_config_filename_format",
-    "/directory/connectors/ldap",
-    "/enterprise/priv_key_path"
+# list of key paths in the root configuration file that will be processed as
+# filenames relative to the configuration's file path
+ROOT_CONFIG_PATH_KEYS = [
+        { 'path':'/dashboard/owning', 'default': DEFAULT_DASHBOARD_OWNING_CONFIG_FILENAME },
+        '/dashboard/owning/enterprise/priv_key_path',
+        '/dashboard/accessors',
```
adobeDan: Should be /dashboard/accessors/*, now that you are handling that properly.

user_sync/config.py
```diff
@@ -694,6 +611,138 @@ def describe_unused_values(self):
...
+                key = keys[level]
+                # if a wildcard is specified at this level, that means we
+                # should process all keys as path values
+                if key == "*":
+                    for key in dictionary.keys():
+                        process_path_key_value(key)
+                elif dictionary.has_key(key):
+                    process_path_key_value(key)
+                # key was not found, but default value was set, so apply it
+                elif default_val:
```
adobeDan: Nice!  Very elegant!

user_sync/config.py
```diff
@@ -694,6 +611,138 @@ def describe_unused_values(self):
...
+                            
+                elif dictionary.has_key(key):
+                    # if the key refers to adictionary, recurse into it to go
+                    # further down the path key
+                    if isinstance(dictionary[key],dict):
+                        process_path_key(dictionary[key], keys, level+1, default_val)
+                        
+                # if the key was not found, but a default value is specified,
+                # drill down further to set the default value
+                elif default_val:
```
adobeDan: brilliant!

PR #104, opened 2017-03-23
improvements to configuration path tests and docs #30

ianmak:
- added support for loading test configuration from files
- reads results from file as well
- updates to documentation


adobeDan:
@ianmak I'll try to review this today; can you be sure you've rebased or merged all the latest rc2 changes to your branch so the PR is as minimal as possible?  Let me know when that's happened, thanks!

ianmak:
I tried both rebase and github desktop merge and it's still left a trail! But all the latest changes are merged in now.

docs/user-manual/index.md
```diff
@@ -351,14 +351,13 @@ support for the User Sync tool.
...
-folder as the command-line executable.
+configuration files.
 
 | Configuration File | Purpose |
 |:------|:---------|
-| dashboard&#x2011;config.yml&nbsp;&nbsp; | Required. Contains credentials and access information for calling the Adobe User Management API. |
-| connector-ldap.yml | Required. Contains credentials and access information for accessing the enterprise LDAP directory. |
-| user-sync-config.yml | Required. Contains configuration options that define the mapping of directory groups to Adobe product configurations and user groups, and that control the update behavior. |
+| user-sync-config.yml | Required. Contains configuration options that define the mapping of directory groups to Adobe product configurations and user groups, and that control the update behavior. You can specify a different configuration file using the `user-sync -c other-config.yml` command option. |
+| dashboard-config.yml | Required. Contains credentials and access information for calling the Adobe User Management API. You can specify a different dashboard configuration file, or directly specify the dashboard configuration parameters in the main configuration file. |
```
adobeDan: You need to make it clear that you can specify a different configuration file for the dashboard as a setting in the main configuration file (rather than as a command-line argument) and refer them to the section on configuration file settings.  Also, don't mention specifying the dashboard parameters directly at this point, because it's an insecure option that's not recommended.

docs/user-manual/index.md
```diff
@@ -351,14 +351,13 @@ support for the User Sync tool.
...
+| connector-ldap.yml | Required. Contains credentials and access information for accessing the enterprise LDAP directory. You can specify a different LDAP configuration file, or directly specify the LDAP configuration parameters in the main configuration file. |
```
adobeDan: Same comment as on the line before.

examples/config files - basic/1 user-sync-config.yml
```diff
@@ -3,7 +3,8 @@ dashboard:
   # By default, it would look for dashboard-owning-config.yml and
   # dashboard-accessor-*-config.yml in the configuration path,
   # with the yml's identifying the owning organization and accessor organizations
-  # respectively.
+  # respectively. All file references are relative to the folder of the
```
adobeDan: Rephrase as: "All relative file paths in any configuration file are taken as relative to that configuration file."

tests/config_file_test.py
```diff
@@ -39,128 +39,109 @@ def assert_eq(self, result, expected, error_message):
...
+    
+            for key in exp.keys():
+                assert_sub_value_eq(exp[key], res[key])
+            
+        try:
+            assert_sub_dict_eq(exp, res)
+        except AssertionError as e:
+            raise AssertionError('%s\n%s' % (err_msg, e.message))
+        
+    def test_load_root_config(self):
```
adobeDan: It would be nice to have just one function that does the test against any pair of files, or even a pair of "file groups" (a set of related config files - master and sub, and a set of "correct answers") where not all the files in a group would need to be in the same directory.  I think, rather than explicitly specifying the paths to convert in the function itself, you could structure the "correct answer" file so that  it had a marker in its keys (or its values), for example, file: as a prefix on the key or value, and the presence of that marker means the key/value holds a relativized form.  Then the test function would use those markers to find the keys/values in the answer, convert them to standard form, and then the answer file would be directly compared against the source file.  (Basically, you would take the source file and copy it, then mark some of its keys and fix the values of those keys, and that would be the answer file.)

tests/config_file_test.py
```diff
@@ -39,128 +39,109 @@ def assert_eq(self, result, expected, error_message):
...
-            self.assert_eq(yml['other']['test-string'], 'test string value should not change', '/other/test-string value should not change')
-            self.assert_eq(yml['other']['test-dict']['test-string-2'], 'this should not change as well', '/other/test-dict/test-string-2 value should not change')
-            self.assert_eq(yml['other']['test-list'][0], 'item-1', '/other/test-list/[0] value should not change')
-            self.assert_eq(yml['other']['test-list'][2]['test-string-3'], 'xyz', '/other/test-list/[2] value should not change')
+        self.assert_dict_eq(yml, exp_yml, 'test root configuration did not match expected configuration')
 
-    @mock.patch('__builtin__.open')
-    @mock.patch('yaml.load')
-    def test_load_root_default_config(self, mock_yaml, mock_open):
+    def test_load_root_default_config(self):
```
adobeDan: As per the comment above, then, this function would go away and be replaced with a single test-invoker that ran the test function over a specified list of file groups.  This is similar to the approach taken now to the default identity test stuff.

examples/config files - custom attributes and mappings/1 user-sync-config.yml
```diff
@@ -3,7 +3,8 @@ dashboard:
   # By default, it would look for dashboard-owning-config.yml and
   # dashboard-accessor-*-config.yml in the configuration path,
   # with the yml's identifying the owning organization and accessor organizations
-  # respectively.
+  # respectively. All file references are relative to the folder of the
```
adobeDan: You missed this one.  Rephrase as "All relative file paths in any configuration file are taken as relative to that configuration file."

adobeDan:
@ianmak, we have moved off of v1 and onto v2 as our working branch.  I have made a LOT of changes in the relative path code to handle the change configuration settings.  I am also making a lot of changes in the testing framework, but I believe much of your stuff will survive that.  I am going to close the PR since it's against v1, which is now a branch we are abandoning (r1 has the actual 1.1.1 release with docs in case we need to fix bugs, and all the after-1.1 changes are in v2).  Can you wait a day or two until the dust settles on the v2 branch and then try rebasing your changes on top of that, or maybe pulling a new branch from v2 and integrating your changes into that?  Thanks!

PR #107, opened 2017-03-24
Initial version of Okta sync functionality

adorton-adobe:
This is the first version of the Okta sync feature.  Refer to the "Okta User Sync" section in the User Manual for setup details.

Done:

* Okta connector module
* Okta config file support
* Dynamic assignment of connector module based on config key in `user-sync-config.yml`
* Extended attribute and group support
* After-map hook support
* Documentation (added to end of User Manual - we can move to a different section or a separate doc if desired)
* Customized Okta Python SDK - needed for country code and extended attribute support (these are not provided by default - see https://github.com/okta/oktasdk-python/issues/8).  The custom-built SDK has been added as a wheel in `misc/build/okta`

Todo:

* Unit tests
* Testing against production-like Okta environment


adobeDan:
Hey, @adorton-adobe, @bhunut-adobe, really great to see this come in!  As we are just closing off the GM work for 1.2 I don't anticipate being @phil-levy  or I being able to do an in-depth review until after that release (which will probably be on Monday or Tuesday).  Does that work for you?

adorton-adobe:
That sounds good to me.

user_sync/connector/directory_okta.py
```diff
@@ -112,7 +112,7 @@ def load_users_and_groups(self, groups, extended_attributes):
         is_using_source_filter = False
         source_filters = options['source_filters']
         source_filter = source_filters.get('all_users_filter')
-        if (source_filter != None):
+        if source_filter is not None:
```
adobeDan: truly idiomatic Python would be:
```
if source_filter:
```
unless you specifically want to include the case where source_filter is an empty list or empty string (which I'm pretty sure you don't).

adorton-adobe: You're right - I had to either commit or discard that change to switch branches (I hate dealing with stashes).  Equality operators should not be used in None comparisons, so I wanted to fix that.

I'm still making changes to the module.  We need to refactor user input workflow.

adorton-adobe:
Ok - we've reworked the user query workflow.  The connector now gets *only* users that are members of mapped groups.  This is something we did not expect to do, but because Okta API queries are slower than LDAP queries (and due to some stakeholder feedback), we decided the change is necessary.

I believe this breaks `--users all` functionality for the Okta connector.  We don't need that functionality at this point.  We can discuss & see what options we have for supporting it, or we can add a note to the Okta section of the doc stating that the Okta connector does not support syncing all users.

We also added pagination since Okta's APIs return a max of 200 user records at a time.

As a side note, I was looking through the user manual today and found that it contains many references to LDAP.  We might want to change that to something more generic like "Directory" or "Sync Source" or something, and only refer to LDAP for matters that concern LDAP specifically.

phil-levy:
I'll look at the "LDAP" question in the documentation.  I actually added many of those since that is a more familiar term for directories in the IT world.

I think it is fine for the okta connector to not support "--users all" but if it doesn't, it should emit an error message if the user users that option.

adorton-adobe:
Thanks, @phil-levy.  Would a warning be OK?  Something like

`WARNING: Okta connector does support "--users all", syncing groups A, B, C`

phil-levy:
For a tool like this, it is usually better to give an error if a parameter is unsupported and get the user to put in something valid so they understand what they are getting with no surprises.  In this case, going from all to a subset of all isn't likely to cause trouble, so this is not as critical.  So I think error is probably preferable but warning is okay.

adobeDan:
@adorton-adobe @bhunut-adobe We have moved off of the v1 branch for active development and all of those changes are now in the v2 (which is our active development branch).  I am going to close this PR for now, and give you some time to absorb the v2 changes and then rebase your okta work onto that.  Once we polish off the v2 release (hopefully end of this next week) we can look to pick up the okta stuff.  Sorry for the inconvenience.

PR #109, opened 2017-03-24
Make some tweaks to architecture diagram

adorton-adobe:
I was asked to make some changes to the architecture diagram.  Here's the new version if you want to replace the version currently used in the User Manual.

Changes:

* Make LDAP arrow one-way to emphasize that we're making a read-only connection
* Add "Read-Only" label to LDAP arrow
* Add HTTPS label to UMAPI arrow
* Increase image resolution
* Tweak upper firewall glyph alignment


adorton-adobe:
I have an Illustrator file that I got from an SVG exported from Gliffy.

I'll submit it in a new PR.  Let me know if any major changes get made to it so I can update our "master" copy in sharepoint.  I'll try to keep an eye on the file.

PR #136, opened 2017-04-03
fix tests and logging for v2rc1

adobeDan:


user_sync/rules.py
```diff
@@ -746,18 +751,18 @@ def is_umapi_user_excluded(self, in_primary_org, user_key, current_groups, do_lo
...
-                    self.logger.info("Excluding adobe user (due to group): %s", user_key)
+                    self.logger.debug("Excluding adobe user (due to group): %s", user_key)
                 self.excluded_user_count += 1
                 return True
             for re in self.exclude_users:
                 if re.match(username):
                     if do_logging:
-                        self.logger.info("Excluding adobe user (due to name): %s", user_key)
+                        self.logger.debug("Excluding adobe user (due to name): %s", user_key)
                     self.excluded_user_count += 1
```
phil-levy: I slightly ambivalent abuot these 3 but let's change it and see if anyone complains.

adobeDan: Yeah, I thought about it for a while, but in fact there are often thousands of these; they are very annoying :).  Let's try it this way.

Issue #110, opened 2017-03-24
Document CSV config-file settings; explain why warning occurs about 'unsused LDAP connector'


raymond-ensemble:
**Steps**
1. Run the user-sync without any arguments (python user-sync.pex) or reading from a csv file only (python user-sync.pex --users file users-file.csv --process-groups)

**Results**
- There would be a warning message: "WARNING config - Found unused keys: ['ldap'] in: <user-sync-config.yml>.directory.connectors"

ianmak:
not sure if this is actually a bug, but I asked Raymond to log this anyways just to discuss it

adobeDan:
I don't actually think this is a bug.  If the config file is set up to use the ldap connector, and the user switches to csv input, I think that's an appropriate warning.  If they are using csv regularly, they should probably replace the ldap settings with csv settings (yes, we have those :).

I'm changing this to be a docs request, because we need to document the CSV settings and show that you can have them in your config file.

adobeDan:
I fixed this in pull #139, although the commit says #4.  This should be closed when that PR is merged, because the docs have also been fixed up.

PR #139, opened 2017-04-04
fix #131 and fix #132: correctly handle default config files

### LINKED ISSUE #4: ldap key not recognized or emits spurrious message
This appears in the example config files but when uncommented result in the following message:

2017-02-01 14:43:53 14292 WARNING config - Found unused keys: ['ldap'] in: <user-sync-config.yml>.directory.connectors

### LINKED ISSUE #131: Empty adobe_groups entry flagged as error (appears in example file)
- directory_group: "Finance"
      adobe_groups:
        # the finance group doesn't use any Adobe products

results in error:
2017-04-03 15:04:49 21744 CRITICAL main - Value not found for key: adobe_groups in: <user-sync-config.yml>.directory_users.groups.[1]

This is directly from the example config file.  The example config file implies this is legal.

Also, the error message isn't great.  Giving group name in addition to [1] would be nice.

### LINKED ISSUE #132: exclude_adobe_groups flagged as error
2017-04-03 15:09:10 5048 ERROR config - Found unused keys: ['exclude_adobe_groups'] in: <user-sync-config.yml>.adobe_users
2017-04-03 15:09:10 5048 CRITICAL main - Detected unused keys that are not ignorable.

This is using the example config file which leaves the list content of this key empty.  If you put something in, it is happy.

adobeDan:
* while I was there, I also decided to fix #4, so now we have no warnings if you define but don't use the ldap or csv connectors.

### DIFF
--- a/user_sync/config.py
+++ b/user_sync/config.py
@@ -127,6 +127,10 @@ class ConfigLoader(object):
         directory_config = self.main_config.get_dict_config('directory_users', True)
         if directory_config != None:
             connectors_config = directory_config.get_dict_config('connectors', True)
+        # make sure neither ldap nor csv connectors get reported as unused
+        if connectors_config:
+            connectors_config.get_list('ldap', True)
+            connectors_config.get_list('csv', True)
         return connectors_config
     
     def get_directory_connector_options(self, connector_name):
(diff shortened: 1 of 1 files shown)


adobeDan:
This also fixes #110, together with the updated config files.

PR #168, opened 2017-04-24
Windows Credential Manager implementation

adobeDan:
Incorporate keyring support.

### DIFF
--- a/examples/config files - basic/3 connector-ldap.yml
+++ b/examples/config files - basic/3 connector-ldap.yml
@@ -30,7 +30,7 @@ credential_manager:
     type: windows_credential_manager
     #service_name is Required for Windows Credential Manger
     service_name: "Internet or Network Address field in Credential Manager"
-    username: "UserName"
+    service_username: "UserName"
 
 # (optional) user_identity_type (default is inherited from main configuration)
 # user_identity_type specifies a default identity type for when directory users
--- a/examples/config files - basic/3 connector-ldap.yml
+++ b/examples/config files - basic/3 connector-ldap.yml
@@ -22,6 +22,16 @@ password: "LDAP password goes here"
 host: "LDAP host URL goes here.  e.g. ldap://ldap.example.com"
 base_dn: "defines the base DN. e.g. DC=example,DC=com"
 
+#Optional
+credential_manager:
+    #This will pull credential from Windows Credential Manager in Control Panel.
+    #UserName and Password field above can be commented out
+    #value: windows_credential_manager
+    type: windows_credential_manager
+    #service_name is Required for Windows Credential Manger
+    service_name: "Internet or Network Address field in Credential Manager"
+    username: "UserName"
+
 # (optional) user_identity_type (default is inherited from main configuration)
 # user_identity_type specifies a default identity type for when directory users
 # are created on the Adobe side (one of adobeID, enterpriseID, federatedID).
--- a/examples/config files - basic/3 connector-ldap.yml
+++ b/examples/config files - basic/3 connector-ldap.yml
@@ -17,20 +17,18 @@
 # You must specify all four of these settings.  Consult with your
 # enterprise directory administrators to get suitable values.
 # These access credentials are sensitive and must be protected.
-username: "LDAP username goes here"
+username: "LDAP or Credential Manager username goes here"
 password: "LDAP password goes here"
 host: "LDAP host URL goes here.  e.g. ldap://ldap.example.com"
 base_dn: "defines the base DN. e.g. DC=example,DC=com"
 
-#Optional
+#(optional)
 credential_manager:
     #This will pull credential from Windows Credential Manager in Control Panel.
-    #UserName and Password field above can be commented out
     #value: windows_credential_manager
     type: windows_credential_manager
     #service_name is Required for Windows Credential Manger
     service_name: "Internet or Network Address field in Credential Manager"
-    service_username: "UserName"
 
 # (optional) user_identity_type (default is inherited from main configuration)
 # user_identity_type specifies a default identity type for when directory users
--- a/user_sync/connector/directory_ldap.py
+++ b/user_sync/connector/directory_ldap.py
@@ -72,6 +72,7 @@ class LDAPDirectoryConnector(object):
         builder.require_string_value('base_dn')
         options = builder.get_options()        
         password = caller_config.get_string('password')
+        caller_config.get_dict("credential_manager", True)
 
         self.user_identity_type = user_sync.identity_type.parse_identity_type(options['user_identity_type'])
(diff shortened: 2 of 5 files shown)


examples/config files - basic/3 connector-ldap.yml
```diff
@@ -22,6 +22,16 @@ password: "LDAP password goes here"
...
 
+#Optional
+credential_manager:
+    #This will pull credential from Windows Credential Manager in Control Panel.
+    #UserName and Password field above can be commented out
+    #value: windows_credential_manager
+    type: windows_credential_manager
+    #service_name is Required for Windows Credential Manger
+    service_name: "Internet or Network Address field in Credential Manager"
+    username: "UserName"
```
adobeDan: I don't get why you are having username be here rather than using the username from above.  That username is available to you in the credentials manager.  If you just want to hide the password, that's fine, but then use username from above.  if you want to hide both the username and the password, that's also fine: just have a username_service and a password_service in the credential_manager section, and have the username be the "password" for the first and the password be the "password" for the second.  But don't be in-between like this.

user_sync/connector/directory_ldap.py
```diff
@@ -72,6 +72,7 @@ def __init__(self, caller_options):
         builder.require_string_value('base_dn')
         options = builder.get_options()        
         password = caller_config.get_string('password')
+        caller_config.get_dict("credential_manager", True)
```
adobeDan: add a comment to explain why you are fetching this value and throwing it away.

user_sync/credential_manager.py
```diff
@@ -42,5 +44,16 @@ def get_credentials(credential_type, credential_id, **kwArgs):
     :type kwArgs: dict
     :rtype dict | str | None
     '''
+    if credential_type == DIRECTORY_CREDENTIAL_TYPE:
+        config = kwArgs['config']
+        if "credential_manager" in config:
+            cred_man = config['credential_manager']
+            for supported in KEYRING_SUPPORTED:
+                if cred_man['type'] == supported:
+                    return get_credentials_from_keyring(cred_man['service_name'], cred_man['username'])
```
adobeDan: you need to wrap this in a try/catch and only return credentials if they were successfully fetched.  Otherwise you should log the failure as Critical and return None (so the original file values are used).  Then the ldap connector will fail to authenticate and raise an exception.

user_sync/credential_manager.py
```diff
@@ -42,5 +44,16 @@ def get_credentials(credential_type, credential_id, **kwArgs):
...
     return None
+
+def get_credentials_from_keyring(service, username):
+    cred = keyring.get_password(service_name=service, username=username)
+    return {'username': username, 'password': cred}
```
adobeDan: see above about username vs. password.

bhunut-adobe:
@adobeDan - Please review, I have modified the code based on your comments.

user_sync/credential_manager.py
```diff
@@ -50,10 +52,16 @@ def get_credentials(credential_type, credential_id, **kwArgs):
...
 
-def get_credentials_from_keyring(service, username):
-    cred = keyring.get_password(service_name=service, username=username)
-    return {'username': username, 'password': cred}
-    
+def get_credentials_from_keyring(service_name, service_username):
+    try:
+        cred = keyring.get_password(service_name=service_name, username=service_username)
+    except Exception as e:
+        raise AssertionException(repr(e))
```
adobeDan: You want to have a wrapper error message here, something like:
```
raise AssertionException("Error accessing platform secure storage: %s " % e)
```

examples/config files - basic/3 connector-ldap.yml
```diff
@@ -30,7 +30,7 @@ credential_manager:
     type: windows_credential_manager
     #service_name is Required for Windows Credential Manger
     service_name: "Internet or Network Address field in Credential Manager"
-    username: "UserName"
+    service_username: "UserName"
```
adobeDan: No, you missed the point of my earlier comment, sorry if I was unclear.  I am saying that, if the customer doesn't mind having the `username` field in plain text in this file, then you should just *remove* the separate specification of username in this section and use the `username` setting that is already specified in the LDAP config.  All you need in this section is the `service_name`.

What confused you might be that I also considered a *separate* use case: If the customer is nervous about even including the username as plain text in the config file, then you could fetch both the username and the password from credential storage.  In that case you have to either have to store a "password" for the `service_name` which is the concatenation of the actual username and password (with some concatenation character), _or_ have two "service name" keys in this section: one that stored the "username" as its password, and one that stored the "password" as its password.  Does that help?

bhunut-adobe: Is there a really security concern of displaying username in plaintext? I demoed this to customer today and they did raise any concern on username part.

the problem with keyring is there is no method that will fetch the username.
get_password() required both service_name and username. This will return is password in a string.

adobeDan: I hope you mean: I demo'd this to the customer today and they did **not** raise any concern on the username part :).

In that case, that's fine, just remove the additional specification of "service_username" from this section and instead use the username field from the original file.  You are given the full config that's in the file as a dictionary, so you can look up the username from there.

bhunut-adobe: oop! that is what I meant... 👍 

I have committed the changes

user_sync/credential_manager.py
```diff
@@ -42,5 +46,22 @@ def get_credentials(credential_type, credential_id, **kwArgs):
     :type kwArgs: dict
     :rtype dict | str | None
     '''
+    if credential_type == DIRECTORY_CREDENTIAL_TYPE:
+        config = kwArgs['config']
+        if "credential_manager" in config:
+            cred_man = config['credential_manager']
+            for supported in KEYRING_SUPPORTED:
+                if cred_man['type'] == supported:
```
adobeDan: If there is no 'type' element in cred_man (user forgot it), this will throw.  Use cred_man.get('type') which will return None if the key is missing.  Even better is to eliminate the loop, and just say
```
if cred_man.get('type') in KEYRING_SUPPORTED:
    return ...
```

user_sync/credential_manager.py
```diff
@@ -42,5 +46,22 @@ def get_credentials(credential_type, credential_id, **kwArgs):
...
+                    return get_credentials_from_keyring(cred_man['service_name'], config['username'])
```
adobeDan: same problem here as above: you can't be guaranteed that 'service_name' is a key in cred_man or that 'username' is a key in config.  So either add some 'in' tests to the if statement above or use the .get(...) form in this call.  I recommend enhancing the if above.

bhunut-adobe: I made the changes as you suggested.

Issue #159, opened 2017-04-08
Allow secure credentials to be retrieved from environment, not disk


adobeDan:
Customer feedback has made it clear that we need to support some more secure options (probably a number of different platform-specific options) for managing credentials.  Since we already have the credential manager hook class for doing this, we should document it clearly and eventually provide some standard implementations that we release as part of the build.

adobeDan:
On AWS, all EC2 instance types support fetching credential files from S3 using IAM roles as a form of secure access.

bhunut-adobe:
@adobeDan I look at Credential Manager class and there isn't much in there. Is it already implemented or it just a placeholder? Documentation on Credential Hook would be awesome. 👍

adobeDan:
@bhunut-adobe It's just a placeholder, although it's compiled.  The proposal is to document it and provide, *perhaps*, one implementation (say via KeyRing) for this release.  Then to provide more implementations in the future.

adobeDan:
After discussing it this morning, @phil-levy and I have decided not to try and do anything without some more input from customers about what's needed and what might work.  In particular, we don't think the credential manager as it stands is even worth documenting; probably we want to replace it with something much clearer and cleaner.  So I am deferring this out of the 2.0 release, which we want to release today, and once we understand what customers need we can do a fast-follow with this enhancement.

@bhunut-adobe, @jameslockman, @adorton-adobe, please note.  We need your help getting clearer requirements for this feature from the field.

bhunut-adobe:
@adobeDan They want to ensure that the password is not in cleartext in the config file. ACL on the file is not secure enough. If the password lives in the config file, it should at least be encrypted and can't be decrypted outside of that machine. The customer also recommending using azure vault credential but this won't be generic enough since most people will be running UST on-prem not azure. 

What would the timeline look like to get the new credential manager hook in place?

adobeDan:
@bhunut-adobe Thanks for the requirements update.  That's from one customer, I gather.  I'll look at some possible approaches and see if @phil-levy and I can agree on something.  We should be able to estimate how long this will take by the end of this week.

jameslockman:
We have other customers. The clear text credential throws a yellow flag on almost every engagement, but conversations with Infosec typically move the engagement along. Every customer wants a more "enterprise" approach to storing this credential, and providing some interface to external encrypted credentials is welcome.

adobeDan:
So here's my proposal:

[...]

PR #172, opened 2017-04-29
intermediate work on #159

adobeDan:
The work on #159 will be in two pieces:

* allowing sensitive data (passwords, client secrets, private key data, etc.) to be stored in the platform's native credential manager; and
* allowing the use of an external process to retrieve configuration files directly into process memory, removing the need to store any sensitive configuration data local to the machine.

This PR is for the first bullet.  It removes the notion of an external credential manager class and instead integrates the use of `keyring` directly into the directory connectors.  It also cleans up a lot of PEP8 style violations and puts a bunch of extra parens in `if` statements and rewrites `None` comparisons into their correct form.

I am submitting this PR to get some code review in advance of the second change, but in addition I wanted to let @phil-levy get started on the documentation.  So I have emailed an encrypted ZIP file to all reviewers (as well as the password) that has configuration files with live (Adobe-internal) credentials in all the different ways you can specify them.


adobeDan:
I've made api_key into a credential as per your review.  Thanks to the refactoring, this was a 1-line change.  I also found and fixed a bug (#173) while I was testing the change.  @phil-levy would appreciate a second review of the later commits.

Issue #167, opened 2017-04-21
--update-user-info and special characters


Gor3t3x:
Hi, 

Last bug i found is using --update-user-info with specials characters like "ç" "é" "è" has a strange behavior.

those characters are pretty common in countries where we speak french...

2017-04-21 14:19:54 2712 INFO processor - Updating info for user key: federatedID,xxx@xxx.xx, changes: {'firstname': 'Andr\xc3\xa9'}
2017-04-21 14:19:54 2712 INFO processor - Updating info for user key: federatedID,xxx@xxx.xx, changes: {'firstname': 'Aur\xc3\xa9lie'}
2017-04-21 14:19:54 2712 INFO processor - Updating info for user key: federatedID,xxx@xxx.xx, changes: {'firstname': 'Andr\xc3\xa9'}
2017-04-21 14:19:54 2712 INFO processor - Updating info for user key: federatedID,xxx@xxx.xx, changes: {'firstname': 'Andr\xc3\xa9'}

Any idea how to resolve this?

Is this related => http://stackoverflow.com/questions/6956799/working-with-unicode-encoded-strings-from-active-directory-via-python-ldap)

EDIT: The problem seems to be present only in console, in adobe dashboard, the special characters are well encoded but still getting this error on each sync with --update-user-info with sames users to modfy

2017-04-21 16:00:08 3368 INFO processor - ---------- Start Sync Umapi --------------------------------
user_sync-2.0-py2-none-any.whl.58f0d6835e0dec629e1283c1839d8f5ad6f21614\user_sync-2.0-py2-none-any.whl\user_sync\rules.py:844:
 UnicodeWarning: Unicode unequal comparison failed to convert both arguments to Unicode - interpreting them as being unequal
  if (value != umapi_value):

ianmak:
This appears to be a limitation of python printing/logging... when you log a dictionary value, sub values containing symbols or non-english characters don't seem to show up properly. I haven't found a quick solution to this, I'll try digging into this a bit more next week.

The Unicode warning you're seeing is sort of a separate problem... I think it's because the directory loads string values, whereas umapi works with unicode strings, and it gets confused when it compares strings with symbols to unicode strings. I'll look into that next week as well...

adobeDan:
[...]

PR #176, opened 2017-05-03
Fix #159 - more secure credential handling

### LINKED ISSUE #159: Allow secure credentials to be retrieved from environment, not disk

### LINKED ISSUE #175: --users value doesn't default to `all` but to `None`
Found while testing the fix of #159: If you run without a `--users` argument, no users are read rather than all ldap users being read.

adobeDan:
Also fixed #175 while I was in there.

### DIFF
--- a/user_sync/app.py
+++ b/user_sync/app.py
@@ -233,25 +233,24 @@ def create_config_loader_options(args):
 
     # --users
     users_args = args.users
-    if users_args is not None:
-        users_action = None if len(users_args) == 0 else user_sync.helper.normalize_string(users_args.pop(0))
-        if (users_action == None or users_action == 'all'):
-            config_options['directory_connector_module_name'] = 'user_sync.connector.directory_ldap'
-        elif (users_action == 'file'):
-            if len(users_args) == 0:
-                raise AssertionException('Missing file path for --users %s [file_path]' % users_action)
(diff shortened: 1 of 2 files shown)


adobeDan:
@phil-levy I believe we are now code-complete for the 2.1 release.  We need docs, to bump the version number, to update the version history file, and to add some example files showing how to use the new feature.  I leave those to you; I've attached a [user-sync-config.yml](https://github.com/adobe-apiplatform/user-sync.py/files/972469/user-sync-config.txt) that shows the process config feature; to make this run on windows, change the `cat` command to a `type` command.  We might want to actually do a sample program that reads the config from a secure S3 and/or Azure file using instance access roles.

adobeDan:
The assumption is that the pipe will return data in the platform-specific charset and the YAML parser will accept that.  It's really between the YAML parser and the process as to what platform-specific charsets are understood; the use of bytes gets away from Python having to understand which one is in use.  The customer may have to adjust what their binary returns if they see inconsistencies.

I need to take a look at #167 to see what's happening there, but it shouldn't relate to this.

PR #178, opened 2017-05-05
Fix #167: allow non-ascii unicode chars in user and group names

### LINKED ISSUE #159: Allow secure credentials to be retrieved from environment, not disk

### LINKED ISSUE #167: --update-user-info and special characters

### LINKED ISSUE #173: excluded user counts are wrong if adobe-only-user-action is exclude
While testing my fix for #159, I noticed that adobe-only users are not added to the excluded user count even if the `adobe-only-user-action` setting is `exclude`.  So I'm fixing that too.

adobeDan:
With this fix, we are unicode-compliant (even in Python 2.7):
* add a `--config-file-encoding` command-line parameter, so the config files (which contain the group mapping) can contain non-ascii characters.
* add a `string_encoding` setting to the directory and ldap configurations, which controls the encoding of the input values.
* Update the sample config files to show the new settings.
* require v2.4.1 of umapi-client, to ensure that it's unicode-capable and matches its validations to the server-side validations.
* Fix #159 again by catching some edge cases around empty credential files.
* Fix #173 again by actually counting the number of primary org users read from Adobe.

### DIFF
--- a/examples/config files - basic/3 connector-ldap.yml
+++ b/examples/config files - basic/3 connector-ldap.yml
@@ -65,6 +65,14 @@ all_users_filter: "(&(objectClass=user)(objectCategory=person)(!(userAccountCont
 # or this one for OpenLDAP: "(&(|(objectClass=groupOfNames)(objectClass=posixGroup))(cn={group}))"
 group_filter_format: "(&(|(objectCategory=group)(objectClass=groupOfNames)(objectClass=posixGroup))(cn={group}))"
 
+# (optional) string_encoding (default value given below)
+# string_encoding specifies the Unicode string encoding used by the directory.
+# All values retrieved from the directory are converted to Unicode before being
+# sent to or compared with values on the Adobe side, to avoid encoding issues.
+# The value must be a Python codec name or alias, such as 'latin1' or 'utf-8.
+# See https://docs.python.org/2/library/codecs.html#standard-encodings for details.
+#string_encoding: utf-8
+
 # (optional) user_identity_type_format (no default)
 # user_identity_type_format specifies how to construct a user's desired identity
 # type on the Adobe side by combining constant strings with attribute values.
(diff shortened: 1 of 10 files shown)


adobeDan:
I use 'rb' mode because non-binary modes are not guaranteed to allow all byte sequences on all platforms; that is, 'rb' is the reliable way to tell the OS not to try to do line buffering or implicit encoding (e.g., some OSes won't allow null characters in non-binary streams).  Since I am going to be doing the decoding explicitly, I need to make sure the OS is not standing in the way.

In fact the YAML parser doesn't care what kinds of line-endings are used; YAML is very explicit about all sequences of line-ending characters (including form feed as well as CR and LF) being interpreted as a single line end.  So in fact doing a binary stream in this context is guaranteed safe because the OS not doing line-end detection can't affect the parse.

In my testing of this feature I am actually using files with both kinds of line end on both platforms.  It works flawlessly, which is very reassuring :).

PR #179, opened 2017-05-05
Prepare for v2.1rc1 release.

adobeDan:
Pushing all the release artifacts to master so we can release from there.


examples/config files - basic/3 connector-ldap.yml
```diff
@@ -17,11 +17,22 @@
 # You must specify all four of these settings.  Consult with your
 # enterprise directory administrators to get suitable values.
 # These access credentials are sensitive and must be protected.
-username: "LDAP username goes here"
+username: "LDAP or Credential Manager username goes here"
```
phil-levy: Doesn't this have to be the LDAP user name?  You then have to use that same name in the credential manager.  If it is the CM user name, where does the LDAP username come from?

PR #187, opened 2017-05-10
Correct links and text

jabogart:
Update for new Authn page locations and terminology


jabogart:
Third time's the charm...
Tested in preview, should be correct now.

PR #190, opened 2017-05-11
Okta Connector Module

adorton-adobe:
With the end of the quarter approaching, I would like to get this reviewed and prepared for integration with the main repo.  Here is a quick summary of changes:

* New Okta directory connector
* Config integration
* Sync tool now dynamically loads connector module based on config key
* New example config
* Docs in user manual (for now - we can move to a separate doc file if needed)
* Unit tests are mostly complete (a few more to come)

@adobeDan and/or @phil-levy - Please take a look and let me know if you have any questions or comments.


adorton-adobe:
@adobeDan or @phil-levy - have either of you had a chance to look at this yet?  Let me know if it would be helpful to set up a review meeting.

adobeDan:
I am (finally) taking a look right now.  Sorry for the delay.

user_sync/connector/directory_okta.py
```diff
@@ -0,0 +1,290 @@
...
+        source_attributes['email'] = user['email'] = profile.email
+
+        source_attributes['identity_type'] = user_identity_type = self.user_identity_type
+        if not user_identity_type:
+            user['identity_type'] = self.user_identity_type
+        else:
+            try:
+                user['identity_type'] = user_sync.identity_type.parse_identity_type(user_identity_type)
+            except user_sync.error.AssertionException as e:
+                self.logger.warning('Skipping user %s: %s', profile.login, e.message)
```
adobeDan: change `e.message` to `e`.  That will cause %s to invoke repr(e) rather than just the message component.  I notice this also needs to be fixed in the LDAP module.  I'll take care of that.

adobeDan:
Hey @adorton-adobe, Phil and I are doing some release planning, and we would love an ETA on the Okta changes so we can figure out if when we could roll it in.

adorton-adobe:
We set an end of may deadline. Functionality is all done barring a few minor unit tests.



Sent from my Verizon, Samsung Galaxy smartphone


-------- Original message --------

bhunut-adobe:
@adorton-adobe Want to split up the work? I can tackle the changes need for directory-okta.py

adorton-adobe:
@adobeDan - I'm not sure what you mean by "Do not attach build products to your PR".  The only wheel we've committed is our modified Okta SDK, which we put in misc/build.  Is there a different way we should include it?  We need the wheel because Kevin had to make some changes to the Okta python SDK to get attributes that aren't provided by default.  Should we just link to his Github fork instead?

user_sync/connector/directory_okta.py
```diff
@@ -0,0 +1,290 @@
...
+    def find_group(self, group):
+        """
+        :type group: str
+        :rtype UserGroup
+        """
+        group = group.strip()
+        options = self.options
+        group_filter_format = options['group_filter_format']
+        try:
+            results = self.groups_client.get_groups(query=group_filter_format.format(group=group))
```
bhunut-adobe: Okta has no concept of a nested group. if a nested AD group gets synced to Okta. Okta will flatten it out.

user_sync/connector/directory_okta.py
```diff
@@ -0,0 +1,290 @@
...
+        except Exception as e:
+            self.logger.warning("Unable to query group")
+            raise user_sync.error.AssertionException(repr(e))
+
+        if results is None:
```
bhunut-adobe: I think if the filter is invalid, okta SDK get_groups method will raise an exception.
Should I just remove try and except from results = self.groups_client.get_groups(query=group_filter_format.format(group=group)) ?

adorton-adobe:
@adobeDan - will you please let me know what you mean by "Do not attach build products to your PR"?  See my previous comment.

adobeDan:
@adorton-adobe I didn't realize that it was a modified SDK build.  I withdraw my comment: having the SDK wheel in there is fine.

user_sync/connector/directory_okta.py
```diff
@@ -58,42 +58,40 @@ class OktaDirectoryConnector(object):
...
+                                 'user.status == "ACTIVE"')
         builder.set_string_value('user_identity_type', None)
-        builder.set_string_value('logger_name', 'connector.' + OktaDirectoryConnector.name)
-        builder.set_dict_value('source_filters', {})
-
-        okta_url = builder.require_string_value('okta_url')
-        if 'https' not in okta_url:
-            okta_url = "https://" + okta_url
+        builder.set_string_value('logger_name', self.name)
+        host = builder.require_string_value('okta_url')
```
adobeDan: Ha!  I meant to change the *configuration setting name* to `host` rather than `okta_url` (as well as the variable name).  Well, I got half of what I wanted.  Is there a reason you are not just calling this `host` since it doesn't have to be a URL?

user_sync/connector/directory_okta.py
adobeDan: No you always want to have an exception wrapper whenever you make a network call.  I just meant this should be a warning rather than a dead stop.  So this looks good.

user_sync/connector/directory_okta.py
```diff
@@ -185,13 +172,10 @@ def iter_group_members(self, group, filter_string, extended_attributes):
...
             #Filtering users based all_users_filter query in config
-            if filter_string:
-                members = self.filter_user(members,filter_string)
-            for member in members:
+            for member in self.filter_users(members, filter_string):
                 profile = member.profile
                 if not profile.email:
-                    # if (last_attribute_name != None):
-                    self.logger.warn('No email attribute for login: %s', profile.login)
+                    self.logger.warning('No email attribute for login: %s', profile.login)
```
adobeDan: by the way, you got this right, and ldap got it wrong - it checks for missing email attribute, but doesn't catch when email is empty.  i'll fix that. :)

user_sync/connector/directory_okta.py
bhunut-adobe: I was waiting for @adorton-adobe to change his part. I kept it this way to make sure UST still working.

adorton-adobe: I've updated the config attribute.

docs/user-manual/index.md
```diff
@@ -1900,7 +1900,7 @@ since Okta uses email address as the primary ID of all users.
 #### Config Example
 
 ```yaml
-okta_url: "Okta URL goes here ie. youcompany.okta.com"
+host: "Okta URL goes here ie. youcompany.okta.com"
```
adobeDan: should be "e.g." (for example) not "ie." (typo for i.e., which means "that is")

examples/config files - basic/5 connector-okta.yml
```diff
@@ -1,4 +1,4 @@
-okta_url: "Your Okta URL goes here ie. youcompany.okta.com"
+host: "Your Okta URL goes here ie. youcompany.okta.com"
```
adobeDan: should be "e.g." (for example) not "ie." (typo for i.e., which means "that is")

bhunut-adobe:
@adobeDan  Regarding --connector argument. Why adding additional argument when user-sync-config.yml directory connector section does the same thing.

why can't we make UST read the key name? currently for LDAP is set to ldap: connector-ldap.yml and for okta it would be okta: connector-okta.yml

adobeDan:
@bhunut-adobe A few reasons:

* I believe customers should be able to prepare a single set of configuration files that establish a consistent group mapping, etc., and then use that same configuration file set against all of their data sources.  Thus they should be able to have the configuration for all supported connectors (currently, ldap, csv, okta) in their single master config file, and choose which connector to use in arguments that are supplied with each run.  What you are proposing would require editing the config file between runs to reorder the connectors, which is both a pain and brittle (because it doesn't allow scheduled tasks that use different connectors to run against the same config).
* The current specification for connector configs is a dictionary, not a list, so there is no innate sense of which connector spec comes "first" in the file.  We would have to change the settings file format to provide an unambiguous mechanism for selecting the "active" connector (and see previous point).
* We have already established a convention wherein command-line args specify which connector to use; to wit, specifying --users file means to select the CSV connector (rather than the default LDAP) and specifying --adobe-only-users-list means don't use any connector.  I'd like not to add a second convention for how connectors are specified.

An alternative to adding a `--connector` argument would be to use the `--users` argument and say that  `--users` is always followed by a connector name and then the specification, as in:

> `-- users ldap` [_args_ ...] / `--users okta` [_args_ ...] / `--users csv` _path_ [_args_ ...]

(where `ldap` is the default and `file` is a synonym for `csv`).  Then you could say, for example,

> `--users okta mapped`
> `--users okta group` _group1_
> `--users okta` (means the same as `--users okta all`)
>
> `--users csv foo.bar mapped` (which we currently can't say, but we should be able to)

I am actually liking that idea a lot more than adding `--connector`, because it's fully backward compatible and actually provides more expressiveness.  Thoughts?

adorton-adobe:
@adobeDan - we can extend users in that way.  I actually started on the connector parameter, however.  Let me see what it will take to update users instead.

To maintain backward compatibility, would `--users [all|file|mapped|group]` still need to work as before?  I assume so...

adobeDan:
@adorton-adobe Yes, we need backward compatibility.  Let me talk to @phil-levy about whether we want `--connector` or to extend `--users`.  I'm guessing he will want `--users` so we get backward compatibility and so that we get the additional functionality of `--users file foobar.csv mapped`

adorton-adobe:
@adobeDan I'm going to stick with `--connector` for now.  argparse gets confused when you include more than two metavar values for some reason.  I'm not sure how to get it to accurately state that there are up to three values for --user (connector type, "mode" and user file).  We can revisit it later.  I need to help a customer deploy this tomorrow.  I'll commit that change and will rebase later this week.  From there, I'll make any outstanding updates and we can review the new changes.

adobeDan:
@adorton-adobe Yeah I like using `--connector` also; let's make sure that when you specify the CSV connector you can use either `file` or `csv` and that it always takes the filename as well.  That way we can do `--connector csv foo.csv --users mapped` which gets us the best of both worlds.

adobeDan:
@phil-levy Now that we have a large customer using this on a regular basis, and others about to go into production, I think we should bring this into the main build before we do any feature changes.  That way the feature changes will be sure to work with Okta as well.  Can you please review and weigh in on the `connector` discussion?

adobeDan:
@adorton-adobe @bhunut-adobe @phil-levy I have merged the latest adorton/user-sync.py Okta branch (as of Mon Oct 30 8am PT) in the adobe-apiplatform/user-sync.py Okta branch.  And I will pull that branch into v2 as soon as we release 2.2.2.  So I am closing this pull request, and request that any further work in the adorton/user-sync.py repo be based on the okta branch of this repo.

Issue #189, opened 2017-05-11
UserSync attempts to remove users from nested product configurations


adobeDan:
There has been some sort of behavior change on the server side, although it appears that it might actually be that the prior behavior was incorrect as per spec and now it is correct.  In any event, the result is that if User Sync maps some users to a user group and other users to a product configuration containing that user group, it then gets server data that shows the users in the user group are also in the product configuration group (which they are *indirectly* but not *directly*), and so User Sync tries to take them out of the product configuration group.  This attempt doesn't do anything, because the users aren't directly in the configuration, but the log is very confusing.

The attached [April 2nd log](https://github.com/adobe-apiplatform/user-sync.py/files/991527/2017-04-02.log.txt) and [May 9th log](https://github.com/adobe-apiplatform/user-sync.py/files/991526/2017-05-10.log.txt) show the behavior differences: the May 9th log shows the additional attempt to remove users from the product configuration.

adobeDan:
It turns out that, although the server API allows controlling whether or not indirect membership is returned, the UMAPI was not exposing it.  So the fix for this bug is to pick up the latest UMAPI release, which provides this control and defaults to direct membership only.

Issue #181, opened 2017-05-10
Add requestID or Tech Account GUID to logs


lesavage:
Can we add the requestID or Tech Account GUID to the user-sync logs? This would help the UMAPI server team with troubleshooting any issues reported by clients.

adobeDan:
@lesavage The tech account GUID is already in the debug log, look for a line containing "UMAPI initialized with options" and look for the value of the `tech_acct` field.

PR #192, opened 2017-05-11
prepare for 2.1rc2

adobeDan:
The base functionality of rc1 is stable, but there have been some server-side behavior changes we need to accommodate (see #189), and we want to eke out a bit more debugging info when requests fail on the server side (see #181).  This build does that.


adobeDan:
@phil-levy @chesterl86 This should be good to build, I am going to merge and tag it.  I suggest starting builds as soon as you get up this morning.

PR #205, opened 2017-05-26
added support for nested ldap groups

### LINKED ISSUE #182: Directory groups can apply to indirect as well as direct members
When following this [guide](https://github.com/adobe-apiplatform/user-sync.py/blob/613cd248720e7501d78c3834367cc9360c833681/docs/user-manual/index.md#working-with-nested-directory-groups-in-active-directory),
UST returned the following error
Multiple LDAP groups found for: AdobeCCETest-All-App

Current Group Structure
```
AdobeCCETest-All-App
	AdobeCCETest-South-Pole
	      jadobe2@adobeccetest.com
```

Log file:
[2017-05-10.txt](https://github.com/adobe-apiplatform/user-sync.py/files/989982/2017-05-10.txt)

ianmak:
- implemented support for expanding nested groups when looking up a group to find it’s members (fixes #182)
- this implementation is not specific to AD or OpenLDAP

### DIFF
--- a/examples/config files - basic/3 connector-ldap.yml
+++ b/examples/config files - basic/3 connector-ldap.yml
@@ -58,6 +58,12 @@ require_tls_cert: False
 # directories might be much simpler: "(&(objectClass=person)(objectClass=top))"
 all_users_filter: "(&(objectClass=user)(objectCategory=person)(!(userAccountControl:1.2.840.113556.1.4.803:=2)))"
 
+# (optional) all_groups_filter (default value given below)
+# all_groups_filter specifies the query used to find all groups in the directory.
+# This is used to build a list of nested users for a given directory group. The value
+# specified here will work with both AD and OpenLDAP.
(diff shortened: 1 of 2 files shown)


adobeDan:
I need to spend some time looking at this.  I think there is a simpler way to do this, and it's not clear to me we want this functionality built in.

ianmak:
Sure thing... Well I went with this initially because I couldn't find an alternative to the 1.2.840.113556.1.4.1941 filter, and David convinced me that LDAP queries are generally pretty fast, so we could use some general filters to resolve nested groups in a platform neutral way theoretically...

adobeDan:
Phil and I have decided to go back to the way we used to do this in v1, that I outlined in my review comments: we will add a setting for doing the LDAP query to find members of a group, and we will make that query for each group we need the members of.  We will only query for all users if we need to fetch all users.  I'm opening a new branch to make this change.

Issue #199, opened 2017-05-19
Credential management doesn't seem to work on Ubuntu or CentOS


raymond-ensemble:
When I tried to run the user-sync with the credential management keyring, I get this error with the keyrings.alt. 

> CRITICAL main - umapi.primary configuration.enterprise: Error accessing secure storage: No recommended backend was available. Install the keyrings.alt package if you want to use the non-recommended backends. See README.rst for details.

When I check the keyrings.alt installation, it says requirement already satisfied.

> pip install keyrings.alt
Requirement already satisfied: keyrings.alt in /usr/lib/python2.7/site-packages
Requirement already satisfied: six in /usr/lib/python2.7/site-packages (from keyrings.alt)

I was just running a normal user-sync 

> ./user-sync --users file users-file.csv --process-groups --adobe-only-user-action exclude

ianmak:
Just to clarify, normally we wouldn't run user-sync with keyrings.alt. We tried it just to see if it would satisfy the tool, but it still complained.

A bit more information:

I tried setting up VM's running desktop versions of centos and ubuntu to see if that would satisfy the keyrings requirement, but it always complains that no recommended backend was available.

I tried looking into this, and I run into the same problem from the python command line using the following lines:

> python
$ import keyring
$ key.get_password(service_name='test_service', username='test_username')
*** No recommended backend was available error ***

I was able to have the system pick up the password values by also installing via pip secretstorage, cffi, dbus-python, it no longer complains that no recommended backend was available and is able to read passwords you see in the "Passwords and Keys" > Login panel.

I tried updating user-sync to also bundle these packages into the app, but it still complains that no recommended backend was available. Though strangely, if I make the keyring.get_password call at the beginning of the app (before even main is called), keyring has no problems retrieving secret service values. So not sure if it's getting confused somewhere between the beginning of the app and at the keyring configuration value loading point or something...

ianmak:
I am also planning to arrange a kubuntu VM so I can try kwallet, but so far we've had no luck retrieving credentials on ubuntu and centos from within user-sync, except in the modified app scenario I noted previously.

adobeDan:
[...]

PR #206, opened 2017-05-30
Adobe apiplatform/bugfix/#199 credential manager ubuntu/centos bug fix

ianmak:
fix for issue #199
- added logic to package additional secure credential python packages on non-windows or non-OS X platforms


adobeDan:
@ianmak Did you investigate whether the packages in setup.py are required in requirements.txt, or whether the wheel command will download them?

ianmak:
ah I did lookup setup.py vs requirements, but misunderstood what they were saying about the alternatives. It looks like pex does try to resolve the required packages. I'll remove requirements.txt and have it wheel the ldap package as needed...

PR #217, opened 2017-06-07
V2 doc update: multi-language and encrypted private key file support

phil-levy:


docs/en/index.md
```diff
@@ -16,14 +20,14 @@ Each time you run User Sync, it looks for differences between the
...
-# Setup and Success Guide
+# [Setup and Success Guide](success-guide/index.md)
 
 The fastest way to get started with User Sync is to read the
 [Setup and Success Guide](success-guide/index.md), which gives
 step-by-step instructions for setting up the needed configuration
 files and running the tool.
 
-# User Manual
+# [User Manual](user-manual/index.md)
```
adobeDan: as above, i suggest no link.

docs/en/index.md
```diff
@@ -16,14 +20,14 @@ Each time you run User Sync, it looks for differences between the
 user and group information in the two systems, and updates the
 Adobe directory to match the information in your directory.
 
-# Setup and Success Guide
+# [Setup and Success Guide](success-guide/index.md)
```
adobeDan: It's tempting to have the headers be links, but it really doesn't format well.  I think just let them be headers and keep the links in the text.

docs/en/index.md
```diff
@@ -32,7 +36,14 @@ point for customizing the behavior of User Sync, as it includes
 instructions for doing custom mappings between customer directory
 information and data on the Adobe side.
 
-# Frequently Asked Questions  
+# [Frequently Asked Questions](FAQ/index.md)
```
adobeDan: as above, I suggest no link.

docs/en/index.md
```diff
@@ -32,7 +36,14 @@ point for customizing the behavior of User Sync, as it includes
 instructions for doing custom mappings between customer directory
 information and data on the Adobe side.
 
-# Frequently Asked Questions  
+# [Frequently Asked Questions](FAQ/index.md) 
 
 We have compiled a [FAQ document](FAQ/index.md) that answers many questions we have been asked and others we expected to be asked.
 
+# Translations Available
```
adobeDan: I think you should also have a sidebar link at the bottom with all translations shown (including English) as either two-letter country codes or fully-localized text, like:
EN | DE | FR | JP | ES
or
English | Deutsch | Français | 日本語 | Español
where each two-letter combo or word is a link to the about.

docs/en/index.md
```diff
@@ -32,7 +36,14 @@ point for customizing the behavior of User Sync, as it includes
...
+
+* [English](../en)
```
adobeDan: No point in having English in the English section.

docs/en/success-guide/index.md
```diff
@@ -2,6 +2,10 @@
 layout: page
 title: Setup and Success Guide
 advertise: Setup and Success Guide
+lang: en
+nav_link: Setup Guide Contents
+nav_level: 1
+nav_order: 100
 ---
 
 Version 1.1, released 2017-03-06
```
adobeDan: should be Version 2.1.1, released 2017-06-09

docs/en/user-manual/command_parameters.md
```diff
@@ -0,0 +1,44 @@
...
+execute the following command in a command shell or from a
+script:
+
+`user-sync` \[ _optional parameters_ \]
+
+The tool accepts optional parameters that determine its
+specific behavior in various situations.
+
+
+| Parameters&nbsp;and&nbsp;argument&nbsp;specifications | Description |
```
adobeDan: It would be great to replace this table with "dictionary list" style text.  Something like:
```
-h
--help
```
Show this help message and exit.

```
--users all
--users file <input_path>
--users group "<group1>,<group2>,..."
--users mapped
```
Specify the users to be selected for sync. The default is `all` meaning all users found in the directory. Specifying `file` means to take input user specifications from the CSV file named by the argument. Specifying `group` interprets the argument as a comma-separated list of groups in the enterprise directory, and only users in those groups are selected. Specifying `mapped` is the same as specifying `group` with all groups listed in the group mapping in the configuration file. This is a very common case where just the users in mapped groups are to be synced.

... and so on.

docs/en/user-manual/deployment_best_practices.md
```diff
@@ -0,0 +1,270 @@
...
+	enterprise:
+	  org_id: your org id
+	  secure_api_key_key: umapi_api_key
+	  secure_client_secret_key: umapi_client_secret
+	  tech_acct: your tech account@techacct.adobe.com
+	  secure_priv_key_data_key: umapi_private_key_data
+
+Note the change of `api_key`, `client_secret`, and `priv_key_path` to `secure_api_key_key`, `secure_client_secret_key`, and `secure_priv_key_data_key`, respectively.  These alternate configuration values give the key names to be looked up in the user keychain (or the equivalent service on other platforms) to retrieve the actual credential values.  In this example, the credential key names are `umapi_api_key`, `umapi_client_secret`, and `umapi_private_key_data`.
+
+The contents of the private key file is used as the value of `umapi_private_key_data` in the credential store.  This can only be done on platforms other than Windows.  See below for how to secure the
```
adobeDan: You might want to just skip describing private key data here and instead always recommend encryption.  And the encryption using PKCS#8 format doesn't have to have a long and complicated passphrase, because it can't be brute forced the way PKCS#5 can.

docs/en/user-manual/deployment_best_practices.md
```diff
@@ -0,0 +1,270 @@
...
+runs the command.  The standard output from the command is captured and that
+output is used as the umapi or ldap configuration file.
+
+The command is run with the working directory as the directory containing the configuration file.
+
+If the command terminates abnormally, User Sync will terminate with an error.
+
+The command can reference a new or existing program or a script.
+
+Note: If you use this technique for the connector-umapi.yml file, you will want to embed the private key data in connector-umapi-yml directly by using the priv_key_data key and the private key value.  If you use the priv_key_path and the filename containing the private key, you would also need to store the private key somewhere
```
adobeDan: Good, this is where the priv_key_data setting is really useful.  You might want to move the above description down here.

docs/en/about.md
```diff
@@ -1,10 +1,13 @@
 ---
-layout: page
+layout: default
 title: About
-permalink: /about/
-advertise: About
+nav_link: About User Sync
```
adobeDan: Remove this file completely.  This Nav Link is actually used for docs/en/index.md, so the About file is never shown.

docs/en/index.md
```diff
@@ -1,9 +1,13 @@
...
-title: User Sync Documentation
+title: About User Sync
+nav_link: About User Sync
+nav_level: 1
+nav_order: 1
+lang: en
 ---
 
-Welcome to the documentation center for [User Sync from Adobe](https://github.com/adobe-apiplatform/user-sync.py/releases/latest).
+Welcome to the documentation center for [User Sync from Adobe](https://github.com/adobe-apiplatform/user-sync.py).
```
adobeDan: This is the right (and only) place to put the version info. Rather than having a line in the User Manual, and one in the Success Guide, just add a paragraph here that says:

Documentation Version: 2.1.1, released 2017-06-09

docs/en/success-guide/index.md
adobeDan: just remove this line, and put the version only in the en/index.md (About) file.

docs/en/success-guide/index.md
```diff
@@ -3,12 +3,12 @@ layout: page
...
 advertise: Setup and Success Guide
 lang: en
-nav_link: Setup Guide Contents
+nav_link: Setup and Success Guide
 nav_level: 1
 nav_order: 100
 ---
 
-Version 1.1, released 2017-03-06
+Version 2.1.1, released 2017-06-09
```
adobeDan: Don't like my idea of moving this to the "About" page, eh?  Sigh, I can live with it :).

docs/en/user-manual/command_parameters.md
```diff
@@ -36,7 +36,8 @@ specific behavior in various situations.
...
+| `--adobe-only-user-list` _filename_ | Specifies a file from which a list of users will be read.  This list is used as the definitive list of "Adobe only" user accounts to be acted upon.  One of the `--adobe-only-user-action` directives must also be specified and its action will be applied to user accounts in the list.  The `--users` option is disallowed if this option is present: only account removal actions can be processed.  |
+| `--string-encoding` _encoding_name_ | Optional.  Specifies the character encoding for the contents of the configuration files themselves.  This includes the main configuration file, "user-sync-config.yml" as well as other configuration files it may reference.  Default is `ascii`.  Character encoding in csv files that are read is controlled by the string_encoding configuration file item, for example, `string_encoding: big5`.  |
```
adobeDan: I think I already gave you my update on this one.

docs/en/user-manual/deployment_best_practices.md
```diff
@@ -93,7 +93,7 @@ Note the change of `api_key`, `client_secret`, and `priv_key_path` to `secure_ap
 The contents of the private key file is used as the value of `umapi_private_key_data` in the credential store.  This can only be done on platforms other than Windows.  See below for how to secure the
 private key file on Windows.
 
-The credential values will be looked up using the specified key names with the user being the org_id value.
+The credential values will be looked up in the secure store using org_id as the username value and the key names in the config file as the key name.
```
adobeDan: should be "settings values in the config file as the key name" (not "key names in the config file")

phil-levy:
Should be good now.  See what you think.

docs/en/user-manual/index.md
```diff
@@ -0,0 +1,120 @@
...
+layout: page
+title: User Manual
+advertise: User Manual
+lang: en
+nav_link: User Manual
+nav_level: 1
+nav_order: 10
+---
+
+Version 2.1rc3, released 2017-05-22
```
adobeDan: Sorry to be a best, but since you didn't take my advice to move all the versioning info to the About page, you need to update this instance as well to:

Version 2.1.1, released 2017-06-09

which is my 61st birthday, by the way! :)

phil-levy:
Don't want to make that change.  I'd like each doc to have its own version number.  Same with localized versions since they may lag and that is how people could see that they are looking at an older version.

phil-levy:
I also didn't make the change you suggested about moving parts of the secure credential description.  That seems like a patch and now that the feature is more baked, the section really needs major cleanup which hopefully Judy can do.

adobeDan:
I get leaving the version in each doc (you've convinced me that's a good idea), but I don't get why you don't want to update the version in the english user manual so that it's correct.  It's still referring to 2.1rc3, and my comment above gave you the correct version number and date to put there.  You've already updated the content to talk about key encryption, so you should update the version number to 2.1.1 and the date to June 9th!

adobeDan:
I think you've gone to dinner,  so I'm going to take this as is so we can do the rc2 build.  I'll manually update the version string after I've merged the pull.

PR #221, opened 2017-06-08
Redirect from old doc location

phil-levy:


adobeDan:
Closing this as we will take these through v2 instead via #222

PR #228, opened 2017-06-13
Issue 52/user sync should work in Python 3 as well as Python 2

ianmak:
- adapted user-sync to support both python 2 and 3


adobeDan:
Thanks, @ianmak!  Now go on vacation!! :)

Issue #201, opened 2017-05-23
empty email attribute in LDAP causes exception


AndyH010493:
Hello,

I am importing my users but am getting the below error. Would you be able to provide some insight on this please?

2017-05-23 12:31:05 31914 ERROR main - Unhandled exception
Traceback (most recent call last):
  File "/home/adobesync/.pex/install/user_sync-2.1-py2-none-any.whl.5b02240d49b48000b66132a75942ae8ff00ba845/user_sync-2.1-py2-none-any.whl/user_sync/app.py", line 345, in main
    begin_work(config_loader)
  File "/home/adobesync/.pex/install/user_sync-2.1-py2-none-any.whl.5b02240d49b48000b66132a75942ae8ff00ba845/user_sync-2.1-py2-none-any.whl/user_sync/app.py", line 203, in begin_work
    rule_processor.run(directory_groups, directory_connector, umapi_connectors)
  File "/home/adobesync/.pex/install/user_sync-2.1-py2-none-any.whl.5b02240d49b48000b66132a75942ae8ff00ba845/user_sync-2.1-py2-none-any.whl/user_sync/rules.py", line 147, in run
    self.read_desired_user_groups(directory_groups, directory_connector)
  File "/home/adobesync/.pex/install/user_sync-2.1-py2-none-any.whl.5b02240d49b48000b66132a75942ae8ff00ba845/user_sync-2.1-py2-none-any.whl/user_sync/rules.py", line 293, in read_desired_user_groups
    if not self.is_selected_user_key(user_key):
  File "/home/adobesync/.pex/install/user_sync-2.1-py2-none-any.whl.5b02240d49b48000b66132a75942ae8ff00ba845/user_sync-2.1-py2-none-any.whl/user_sync/rules.py", line 406, in is_selected_user_key
    username = self.get_username_from_user_key(user_key)
  File "/home/adobesync/.pex/install/user_sync-2.1-py2-none-any.whl.5b02240d49b48000b66132a75942ae8ff00ba845/user_sync-2.1-py2-none-any.whl/user_sync/rules.py", line 901, in get_username_from_user_key
    return self.parse_user_key(user_key)[1]
  File "/home/adobesync/.pex/install/user_sync-2.1-py2-none-any.whl.5b02240d49b48000b66132a75942ae8ff00ba845/user_sync-2.1-py2-none-any.whl/user_sync/rules.py", line 898, in parse_user_key
    return user_key.split(',')
[...]

PR #232, opened 2017-06-15
Fix #227: crash on None value for user key

### LINKED ISSUE #227: User-Sync tool crashing with "AttributeError: 'NoneType' object has no attribute 'split'"
Hi Dan,

Strange one. I have installed version 2.1.1 of the user-sync tool but I'm still getting the same error. See below:

2017-06-13 10:10:19 32744 ERROR main - Unhandled exception
Traceback (most recent call last):
File "/home/adobesync/.pex/install/user_sync-2.1.1-py2-none-any.whl.d74436c53f3ebfa40d11c3b3623530ef8a9bb6e8/user_sync-2.1.1-py2-none-any.whl/user_sync/app.py", line 352, in main
begin_work(config_loader)
File "/home/adobesync/.pex/install/user_sync-2.1.1-py2-none-any.whl.d74436c53f3ebfa40d11c3b3623530ef8a9bb6e8/user_sync-2.1.1-py2-none-any.whl/user_sync/app.py", line 210, in begin_work
rule_processor.run(directory_groups, directory_connector, umapi_connectors)
File "/home/adobesync/.pex/install/user_sync-2.1.1-py2-none-any.whl.d74436c53f3ebfa40d11c3b3623530ef8a9bb6e8/user_sync-2.1.1-py2-none-any.whl/user_sync/rules.py", line 154, in run
self.read_desired_user_groups(directory_groups, directory_connector)
File "/home/adobesync/.pex/install/user_sync-2.1.1-py2-none-any.whl.d74436c53f3ebfa40d11c3b3623530ef8a9bb6e8/user_sync-2.1.1-py2-none-any.whl/user_sync/rules.py", line 300, in read_desired_user_groups
if not self.is_selected_user_key(user_key):
File "/home/adobesync/.pex/install/user_sync-2.1.1-py2-none-any.whl.d74436c53f3ebfa40d11c3b3623530ef8a9bb6e8/user_sync-2.1.1-py2-none-any.whl/user_sync/rules.py", line 411, in is_selected_user_key
username = self.get_username_from_user_key(user_key)
File "/home/adobesync/.pex/install/user_sync-2.1.1-py2-none-any.whl.d74436c53f3ebfa40d11c3b3623530ef8a9bb6e8/user_sync-2.1.1-py2-none-any.whl/user_sync/rules.py", line 907, in get_username_from_user_key
return self.parse_user_key(user_key)[1]
File "/home/adobesync/.pex/install/user_sync-2.1.1-py2-none-any.whl.d74436c53f3ebfa40d11c3b3623530ef8a9bb6e8/user_sync-2.1.1-py2-none-any.whl/user_sync/rules.py", line 904, in parse_user_key
return user_key.split(',')
AttributeError: 'NoneType' object has no attribute 'split'
2017-06-13 10:10:19 32744 INFO main - ========== End Run (User Sync version: 2.1.1) (Total time: 0:00:01)

adobeDan:
I thought I had fixed this when I put in the fix for #201, but it looks like I didn't catch all the cases.  So now I'm going directly to the source, looking for empty user keys as they are computed, ignoring those users, and logging their details so we can figure out why their user key is empty.


adobeDan:
So it turns out users with "improperly formatted email addresses" (no domain, illegal characters, etc.) were crashing us. Since this is a rare event on the customer side, I think the general warning message being used (which simply says that the user is being skipped because they don't have a valid user key), along with the dump of the user data, is probably enough for customers to find the bad users.  I've had the fix confirmed already by the reporting customer in the 2.2a1 build I made for him.

PR #238, opened 2017-06-19
fix #236: allow users to be pushed rather than sync'd to Adobe.

### LINKED ISSUE #236: Setting users to a known state shouldn't require fetching Adobe users
If an organization is very, very large, and the customer wants to run a job that creates or updates specific users to have a specific set of known attributes and group memberships (similar to a bulk upload from the admin console), the customer shouldn't have to wait for User Sync to fetch the Adobe-side users (which may take a long time) and to do the sync.  Instead, the customer should be able to "directly populate" the users into the org with those attribute values and group memberships, instructing user sync either to create or update each individual user as necessary.

adobeDan:
Adds a `--strategy` command line argument (default 'sync', other value 'push') that controls whether users are fetched from Adobe and sync'd or not.  In the push case, the users are just blindly created against the Adobe org(s) and (if `--process-groups` is specified) any mapped groups are added to them.  This will be idempotent for users that exist on the Adobe side, and will not remove any mapped groups (only add them).


adobeDan:
@phil-levy Let's discuss whether we want to also remove mapped-but-not-assigned groups from the pushed users.

user_sync/rules.py
```diff
@@ -154,15 +162,14 @@ def run(self, directory_groups, directory_connector, umapi_connectors):
             load_directory_stats.log_start(logger)
             self.read_desired_user_groups(directory_groups, directory_connector)
             load_directory_stats.log_end(logger)
-            should_sync_umapi_users = True
-        else:
-            # no directory users to sync with
-            should_sync_umapi_users = False
 
-        umapi_stats = JobStats("Sync Umapi", divider="-")
+        umapi_stats = JobStats('Sync Umapi' if self.sync_umapi else 'Push to Umapi', divider="-")
```
phil-levy: This isn't a big issue, but I think we write UMAPI all caps in most places, including the documentation.  In addition, is that the term we use as the destination or is it more "Adobe" or "Adobe User Management".

user_sync/rules.py
```diff
@@ -408,6 +415,20 @@ def process_umapi_users(self, umapi_connectors):
                 self.logger.critical("Shouldn't happen! In secondary umapi %s, the following users were not found: %s",
                                      umapi_name, secondary_updates_by_user_key.keys())
 
+    def push_umapi_users(self, umapi_connectors):
+        """
+        This is where we push directory users to the Adobe side "as is".
+        :type umapi_connectors: UmapiConnectors
+        """
+        if umapi_connectors.get_secondary_connectors():
+            self.logger.debug('Pushing users to primary umapi...')
```
phil-levy: UMAPI

user_sync/rules.py
```diff
@@ -408,6 +415,20 @@ def process_umapi_users(self, umapi_connectors):
...
+        else:
+            self.logger.debug('Pushing users to umapi...')
```
phil-levy: ditto

user_sync/rules.py
adobeDan: Ha!  Great minds think alike.  I originally changes this to UMAPI but then I thought "Oh, no, Phil will kill me if I make a cosmetic change in the reports, because he will have to redo the docs and will worry that there are customers searching on that string!" :)  I am HAPPY to change this occurrence and will do so right now!

user_sync/rules.py
adobeDan: This is a debug log message and I intentionally chose "umapi" lower-case so that it will align with the lower-case in the config file and also the other debug logging messages where "umapi" (lower-case) is part of the log string.  So I'm not willing to change this one.

user_sync/rules.py
adobeDan: Not willing to change for same reasons as above.

adobeDan:
@phil-levy I think this is ready to go.  Please review and let me know; if you like it I'll merge tonight and build the rc1 on various platforms.

PR #279, opened 2017-09-25
Update connector-ldap.yml

Luci2015:
when all_users_filter and all_groups are being used

### DIFF
--- a/examples/config files - basic/3 connector-ldap.yml
+++ b/examples/config files - basic/3 connector-ldap.yml
@@ -55,10 +55,10 @@ require_tls_cert: False
 # The default value specified below is appropriate for Active Directory, which has a
 # special field that is used to enable and disable users.  The value for OpenLDAP
 # directories might be much simpler: "(&(objectClass=person)(objectClass=top))"
-# In order to obtain a more fine tunned list of users, adapt the query below to your needs.
-# This filter can work by itself if "--users all" is used (all obtained users will make the
-# object of synchronisation, minus the excluded one) OR it creates the base users for further 
-# filtering via group filters
+# In order to obtain a more fine-tunned list of users, adapt the query below to your needs.
+# The result of this filter predicate is the base user list considered by the User Sync tool;
+# that is, all LDAP queries issued by User Sync include this predicate to filter out LDAP resources
+# that are not users to be synchronized.
 all_users_filter: "(&(objectClass=user)(objectCategory=person)(!(userAccountControl:1.2.840.113556.1.4.803:=2)))"
 
 # (optional) group_filter_format (default value given below)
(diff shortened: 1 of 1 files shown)


Luci2015:
I left out --process-groups intentionally because it does not seem to process changes just for the group you are setting filter, but for all mapped groups.
I agree that there is a default all user filter that runs even if I comment the one in config file; what I meant was that if enabled it becomes the new “all users that AD can offer and that are getting attention form the tool, syncing to Adobe side accordingly")

examples/config files - basic/3 connector-ldap.yml
```diff
@@ -52,10 +52,13 @@ search_page_size: 200
 require_tls_cert: False
 
 # (optional) all_users_filter (default value given below)
-# all_users_filter specifies the query used to find all users in the directory.
-# The default value specified here is appropriate for Active Directory, which has a
+# The default value specified below is appropriate for Active Directory, which has a
 # special field that is used to enable and disable users.  The value for OpenLDAP
 # directories might be much simpler: "(&(objectClass=person)(objectClass=top))"
+# In order to obtain a more fine tunned list of users, adapt the query below to your needs.
```
adobeDan: "fine tunned" -> fine-tuned

examples/config files - basic/3 connector-ldap.yml
```diff
@@ -52,10 +52,13 @@ search_page_size: 200
...
+# This filter can work by itself if "--users all" is used (all obtained users will make the
+# object of synchronisation, minus the excluded one) OR it creates the base users for further 
+# filtering via group filters
```
adobeDan: How about changing the above three lines to:
Only users who are accepted by this filter predicate are ever considered by the User Sync tool;
that is, all LDAP queries issued by User Sync include this predicate to filter out LDAP resources
that are not users to be synchronized.

examples/config files - basic/3 connector-ldap.yml
```diff
@@ -67,6 +70,7 @@ all_users_filter: "(&(objectClass=user)(objectCategory=person)(!(userAccountCont
 # You will likely want to replace it with a simpler query customized for your directory,
 # such as this one for Active Directory: "(&(objectCategory=group)(cn={group}))"
 # or this one for OpenLDAP: "(&(|(objectClass=groupOfNames)(objectClass=posixGroup))(cn={group}))"
+# If enabled, will be used whenever "--users group" or "--users mapped" arguments are used
```
adobeDan: This is very misleading.  Remove this lie completely.  This predicate is used whenever we need the contents of a group, whether specified in the command line or specified in the group mapping.

examples/config files - basic/3 connector-ldap.yml
```diff
@@ -77,6 +81,9 @@ group_filter_format: "(&(|(objectCategory=group)(objectClass=groupOfNames)(objec
 # that is contained in the group.  If you want indirect containment, then
 # use this value instead of the default:
 # group_member_filter_format: "(memberOf:1.2.840.113556.1.4.1941:={group_dn})"
+# If enabled, will be used whenever "--users group" or "--users mapped" arguments 
+# are used and will filter more the result of all_users_filter - if set,
+# otherwise all users from directory's base_dn would make it's base.
```
adobeDan: Again, this is misleading and should be removed.  We use this whenever we need the content of a group, whether specified in the command line or in the group mapping.  I believe the only confusing documentation in this file is the description of the all_users_filter; see above.

Luci2015:
will use v2 branch

PR #280, opened 2017-09-26
Update connector-ldap.yml

Luci2015:
included some more comments for all_users_filter

### DIFF
--- a/examples/config files - basic/3 connector-ldap.yml
+++ b/examples/config files - basic/3 connector-ldap.yml
@@ -52,7 +52,10 @@ search_page_size: 200
 require_tls_cert: False
 
 # (optional) all_users_filter (default value given below)
-# all_users_filter specifies the query used to find all users in the directory.
+# In order to obtain a more fine-tunned list of LDAP users,  adapt the filter to your needs.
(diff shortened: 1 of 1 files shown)


examples/config files - basic/3 connector-ldap.yml
```diff
@@ -52,7 +52,10 @@ search_page_size: 200
 require_tls_cert: False
 
 # (optional) all_users_filter (default value given below)
-# all_users_filter specifies the query used to find all users in the directory.
+# In order to obtain a more fine-tunned list of LDAP users,  adapt the filter to your needs.
+# Only resulted users from this filter predicate are ever considered by the User Sync tool;
+# that is, all LDAP queries issued by User Sync include this predicate to filter out LDAP resources
+# that are not subject of synchronization.
```
adobeDan: Your wording is not idiomatic english and needs to be changed.  How about:

Use this filter to control exactly which LDAP resources are considered for synchronization.
If a user is not accepted by this filter he will never be seen by the User Sync tool, even if
your command line specifies `--users all` or he is in a group you have specified in your
command line (`--users group g1`) or in your configuration file (for group mapping).

PR #284, opened 2017-10-11
Add feature that allows users to define the log file name.

cgravatt:
I want to use log rotate to handle log rotation for log files of the usersync tool, but log rotate is not very compatible with log files where the log filename is changed daily. I've added an option to allow you to specify a log name or maintain current adobe implementation.

Please accept or deny my proposed changes as deemed appropriate.

### DIFF
--- a/examples/config files - basic/1 user-sync-config.yml
+++ b/examples/config files - basic/1 user-sync-config.yml
@@ -217,6 +217,13 @@ logging:
   # If a relative path is specified, it is interpreted relative to this configuration
   # file, not relative to the User Sync process working directory.
   file_log_directory: logs
+  
+  # (optional) file_log_name (default value "default")
+  # The default value will log to a file with the following format "YYYY-MM-DD.log", and
+  # all logging for runs started that day (local time) are placed in that file.
+  # However, this option can be used to always log to a single file i.e. adobesync.log
+  # This method would be preferred for users wanting to use logrotate to manage the log files
+  file_log_name: default
 
   # (optional) file_log_level (default value "info")
   # This determines the detail level of the information logged to files.
--- a/examples/config files - basic/1 user-sync-config.yml
+++ b/examples/config files - basic/1 user-sync-config.yml
@@ -212,8 +212,6 @@ logging:
 
   # (optional) file_log_directory (default value "logs")
   # An absolute or relative path to the directory where log files should be placed.
-  # A single file is created per day, named with format "YYYY-MM-DD.log", and all
-  # logging for runs started on that day (local time) are placed in that file.
   # If a relative path is specified, it is interpreted relative to this configuration
   # file, not relative to the User Sync process working directory.
   file_log_directory: logs
--- a/user_sync/app.py
+++ b/user_sync/app.py
@@ -121,6 +121,7 @@ def init_log(logging_config):
     builder.set_bool_value('log_to_file', False)
     builder.set_string_value('file_log_directory', 'logs')
     builder.set_string_value('file_log_level', 'info')
+    builder.set_string_value('file_log_name', 'default')
     builder.set_string_value('console_log_level', 'info')
     options = builder.get_options()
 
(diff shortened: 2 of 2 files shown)


adobeDan:
@cgravatt Can you please open an issue that describes the desired functionality?  It's hard to evaluate your pull request without a spec for what it's supposed to do.

adobeDan:
@chesterl86 @cgravatt I have written up #292 as my understanding of what's desired here.  Please take a look and let me know if that's what you mean.  Assuming it is, I'll review the pull request.  I'd like to get this into the 2.2.2 release if possible.

examples/config files - basic/1 user-sync-config.yml
```diff
@@ -212,11 +212,16 @@ logging:
 
   # (optional) file_log_directory (default value "logs")
   # An absolute or relative path to the directory where log files should be placed.
-  # A single file is created per day, named with format "YYYY-MM-DD.log", and all
-  # logging for runs started on that day (local time) are placed in that file.
   # If a relative path is specified, it is interpreted relative to this configuration
   # file, not relative to the User Sync process working directory.
   file_log_directory: logs
+  
+  # (optional) file_log_name (default value "default")
```
adobeDan: The name of this parameter  should be `file_log_name_format`.

The value of this parameter should be a Python format string that can refer to a single positional argument which is a `datetime.datetime` object produced by calling `datetime.now()` at the start of the run.

The default value of this string should be `'{%Y-%m-%d}.log'`, which produces the current default value.

user_sync/app.py
```diff
@@ -121,6 +121,7 @@ def init_log(logging_config):
     builder.set_bool_value('log_to_file', False)
     builder.set_string_value('file_log_directory', 'logs')
     builder.set_string_value('file_log_level', 'info')
+    builder.set_string_value('file_log_name', 'default')
```
adobeDan: same change as above: parameter is `file_log_name_format` and default value is `'{%Y-%m-%d}.log'`

user_sync/app.py
```diff
@@ -148,7 +149,13 @@ def init_log(logging_config):
         if not os.path.exists(file_log_directory):
             os.makedirs(file_log_directory)
 
-        file_path = os.path.join(file_log_directory, datetime.date.today().isoformat() + ".log")
+	file_log_name = "{datestamp}.log".format(datestamp=datetime.date.today().isoformat())
```
adobeDan: This line is not necessary.  since you set the `file_log_name_format` in the builder, it will always be present.

user_sync/app.py
```diff
@@ -148,7 +149,13 @@ def init_log(logging_config):
...
+
+	if options['file_log_name']:
```
adobeDan: This line is not necessary.  since you set the `file_log_name_format` in the builder, it will always be present.

user_sync/app.py
```diff
@@ -148,7 +149,13 @@ def init_log(logging_config):
...
+	    if options['file_log_name'] != 'default':
```
adobeDan: It's now a format, so we don't have to test for any given value.

user_sync/app.py
```diff
@@ -148,7 +149,13 @@ def init_log(logging_config):
...
+	        file_log_name = options['file_log_name']
```
adobeDan: This should unconditionally be ```python
file_log_name = option['file_log_name_format'].format(datetime.now())
```

File: user_sync/app.py [hunk 7]
```diff
@@ -340,6 +347,7 @@ def main():
     try:
         try:
             args = process_args()
+
```
adobeDan: PEP8: no blank lines in a try/except sequence.  Also it's very bad practice in a PR to make changes to code that has nothing to do with the code you mean to fix, because it produces unnecessary merge errors.

adobeDan:
This feature was added in 94e1090, so there's no need for this PR anymore.

PR #287, opened 2017-10-25
Escape LDAP special chars

adorton-adobe:
LDAP defines a number of special characters that must be escaped in LDAP queries:

http://www.rfc-editor.org/rfc/rfc4515.txt

The sync tool must escape any of these special characters that may be present in group names, group DNs, or anywhere else we use `string.format()` to inject dynamic values into the UST's LDAP queries.

This patch should resolve that issue.  I've tested it with special characters in group names as well as in OUs that contain groups (which shows up in the group DN).

### DIFF
--- a/user_sync/connector/directory_ldap.py
+++ b/user_sync/connector/directory_ldap.py
@@ -133,7 +133,7 @@ class LDAPDirectoryConnector(object):
             if not group_dn:
                 self.logger.warning("No group found for: %s", group)
                 continue
-            group_member_subfilter = group_member_filter_format.format(group_dn=group_dn)
+            group_member_subfilter = self.format_ldap_query_string(group_member_filter_format, group_dn=group_dn)
             if not group_member_subfilter.startswith('('):
                 group_member_subfilter = six.text_type('(') + group_member_subfilter + six.text_type(')')
             user_subfilter = all_users_filter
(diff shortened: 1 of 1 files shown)


user_sync/connector/directory_ldap.py
```diff
@@ -179,7 +179,7 @@ def find_ldap_group_dn(self, group):
         group_filter_format = six.text_type(options['group_filter_format'])
         try:
             res = connection.search_s(base_dn, ldap.SCOPE_SUBTREE,
-                                      filterstr=group_filter_format.format(group=group), attrsonly=1)
+                                      filterstr=self.format_ldap_query_string(group_filter_format, group=group), attrsonly=1)
```
adobeDan: Please pull out the `filterstr` argument calculation into a separate variable for readability:
```python
filter_string = self.format_ldap_query_string(group_filter_format, group=group)
res = connection.search_s(base_dn, ldap.SCOPE_SUBTREE, filterstr=filter_string, attrsonly=1)
```

user_sync/connector/directory_ldap.py
```diff
@@ -324,6 +324,33 @@ def iter_search_result(self, base_dn, scope, filter_string, attributes):
                 connection.abandon(msgid)
             raise
 
+    @staticmethod
+    def format_ldap_query_string(query, **kwargs):
+        """
+        To be used with any string that will be injected into a LDAP query - this escapes a few special characters that
```
adobeDan: Function comments need to be imperative, as in "Escape LDAP special characters that may appear in injected query strings."  Then put the usage info on the next line: "Should be used with any string that will be injected into a LDAP query."  Also it's "an LDAP" not "a LDAP" :).

user_sync/connector/directory_ldap.py
```diff
@@ -324,6 +324,33 @@ def iter_search_result(self, base_dn, scope, filter_string, attributes):
...
+        may appear in DNs, group names, etc.
+        :param query:
+        :param kwargs:
+        :return:
+        """
+        escape_chars = six.text_type('*()\\&|<>~!:')
```
adobeDan: give the rfc reference for this set in a comment.

PR #311, opened 2017-12-01
Rebase v2 onto feature/test-framework

adorton-adobe:
To add tests for group sync, I needed to rebase `v2` onto `feature/test-framework`.


adobeDan:
So I have attempted this rebase in a local branch, and I realize that I don't like the approach that was taken to mocking the UMAPI connection (by giving a command-line parameter to the user sync tool).  I would rather not have the User Sync tool even be aware that the mocking is going on, and I certainly don't want to expose that capability to customers.

The alternate approach that I would like to propose is based on adobe-apiplatform/umapi-client.py#33, which suggests that environmental arguments be used to control the umapi-client behavior around connections.  That way the test tool can just set the appropriate environment variables and we need to make *no* changes in the user sync tool at all.  (Especially since I've now allowed the User Sync tool to easily be invoked by other Python apps.)

So I am going to close this pull request and instead update the user-sync portions of the test framework to match the current v2.  @adorton-adobe, @bhunut-adobe I can probably get the umapi-client changes made in the next week, but if you want to beat me to it feel free to propose a PR over there.  Otherwise I think you can just focus on getting the test framework to work using environment variables instead of user-sync changes.

Issue #319, opened 2017-12-14
invalid country code (lower case)


Luci2015:
There are cases when "c" attribute comes in as lower case value from source.
Users having the lowercase value for country code do not get created in Admin Console for this reason.
Forcing the value to be uppercase before the API call, might cover this mistake.

PR #320, opened 2017-12-14
enforce country code to be uppercase

Luci2015:
resolving issue #319 
This change is proposed for csv and ldap only.


user_sync/connector/directory_ldap.py
```diff
@@ -269,9 +269,9 @@ def iter_users(self, users_filter, extended_attributes):
             if sn_value is not None:
                 user['lastname'] = sn_value
             c_value = LDAPValueFormatter.get_attribute_value(record, six.text_type('c'))
-            source_attributes['c'] = c_value
+            source_attributes['c'] = c_value.upper() if c_value else None
```
adobeDan: The source_attributes should not be uppercased, because it's supposed to show through the actual source value.

Luci2015:
changes applied; also for Okta

user_sync/connector/directory_okta.py
```diff
@@ -227,9 +227,10 @@ def convert_user(self, record, extended_attributes):
             source_attributes['lastName'] = None
 
         if profile.countryCode:
-            source_attributes['countryCode'] = user['country'] = profile.countryCode
+            source_attributes['countryCode'] = profile.countryCode
+            user['country'] = profile.countryCode.upper()
         else:
-            source_attributes['countryCode'] = None
+            source_attributes['countryCode'] = user['country'] = None
```
adobeDan: No, the field on the user object is not supposed to be set in this case; only the source object.  This line should stay the same.

Luci2015:
Sorry, I just thought it was a missed users['country'] assignment to None since, following the logic for source_attributes in directory_ldap file. I did not follow where these variable go after..

Issue #324, opened 2018-01-04
CRITICAL main - Unexpected LDAP failure reading group members: {'desc': 'Critical extension is unavailable'}


bajwa-adobe:
I am running the UST v2.2.2 against Oracle Directory Enterprise Directory 11g Release 1. However, I am getting following error
 
`2018-01-04 21:25:43 25356 CRITICAL main - Unexpected LDAP failure reading group members: {'desc': 'Critical extension is unavailable'}`

I tried to narrow down the issue that seems to be in paged LDAP query or **directory_ldap.py: iter_search_result** code for Oracle Directory, I have created a script **test_dir2.py.txt** which also report same issue.

The ldapsearch seems to return correct results. Same UST's setup work for AD (with correct AD filters). 

[connector-ldap.yml.txt](https://github.com/adobe-apiplatform/user-sync.py/files/1604814/connector-ldap.yml.txt)
[ldapsearch_result.txt](https://github.com/adobe-apiplatform/user-sync.py/files/1604805/ldapsearch_result.txt)
[2018-01-04_Issue.log](https://github.com/adobe-apiplatform/user-sync.py/files/1604806/2018-01-04_Issue.log)
[test_dir2.py.txt](https://github.com/adobe-apiplatform/user-sync.py/files/1605188/test_dir2.py.txt)

bajwa-adobe:
It seems Oracle Directory Enterprise Edition 11g Release 1 does not support PagedResults which throws this error.

However, the UST seems to return correct results if I comment out paged results, and instead use **connection.search(...)** in **directory_ldap.py: iter_search_result** code

https://docs.oracle.com/cd/E19455-01/806-5580/6jej518p5/index.html
http://www.ietf.org/rfc/rfc2696.txt

[supportedControl.txt](https://github.com/adobe-apiplatform/user-sync.py/files/1608349/supportedControl.txt)

PR #326, opened 2018-01-08
Ability to disable paged results for Enterprise Directory

bajwa-adobe:
If Simple Page Mode Control ( 1.2.840.113556.1.4.319) is not supported.
Use search_page_size: 0 to disable paged results from Enterprise
Directory

#324 UST throws error 'Critical extension is unavailable' for oracle directory that does not support Simple Page Mode Control

### DIFF
--- a/user_sync/connector/directory_ldap.py
+++ b/user_sync/connector/directory_ldap.py
@@ -300,6 +300,7 @@ class LDAPDirectoryConnector(object):
                 msgid = connection.search(base_dn, scope,
                                           filterstr=filter_string, attrlist=attributes)
                 result_type, response_data, _rmsgid = connection.result2(msgid)
+                msgid = None
                 if result_type in self.expected_result_types and (response_data is not None):
                     for item in response_data:
                         yield item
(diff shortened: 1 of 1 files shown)


user_sync/connector/directory_ldap.py
```diff
@@ -294,32 +294,40 @@ def iter_search_result(self, base_dn, scope, filter_string, attributes):
...
-                        lc.cookie = cookie = pctrls[0].cookie
-                        if not cookie:
-                            has_next_page = False
-                if has_next_page:
-                    msgid = connection.search_ext(base_dn, scope,
-                                                  filterstr=filter_string, attrlist=attributes, serverctrls=[lc])
+            if search_page_size == 0:
+                msgid = connection.search(base_dn, scope,
+                                          filterstr=filter_string, attrlist=attributes)
+                result_type, response_data, _rmsgid = connection.result2(msgid)
```
adobeDan: I think you need to set `msgid = None` here after retrieving results to avoid problems at GeneratorExit should one be thrown.  There's no connection to abandon in this case.

PR #368, opened 2018-07-26
Added new parameter: --adobe-users all|mapped|group

bhunut-adobe:
This feature allow UST to pull in All Adobe Users or Only the adobe groups specified in User-Sync-Config.yml or a list of group.

Example:

--adobe-users mapped
--adobe-users all
--adobe-users group "GroupA,GroupB,GroupC"


adorton-adobe:
@bhunut-adobe please add a corresponding issue so we can track this feature

bhunut-adobe:
#371

janssenda:
Totally digging this feature! I have a client for whom I may utilize it for because they have a widely distributed system.   One thing I'm noticing playing with it is that if I am using users mapped and adobe users mapped, and I remove a user from the mapped AD group, they are correctly removed from the adobe group.  However, when re-adding that user to the AD group, the message states that they are being created as a new user on the console side because they are not in view of the tool when it pulls the user list.

BUT, despite the message, it still works - the user is not actually recreated, but they are mapped to the group again.  I think this is expected behavior since the tool is doing exaclty what it should given the context - but maybe we can make a note about this in the doc somewhere?

adorton-adobe:
@bhunut-adobe can you please add some documentation about this in [usage scenarios](https://adobe-apiplatform.github.io/user-sync.py/en/user-manual/usage_scenarios.html) and an entry for the new command-line option in [command parameters](https://adobe-apiplatform.github.io/user-sync.py/en/user-manual/command_parameters.html)?

We'll also need something about it in the revised invocation_defaults section, but I still need to work on that.

In the meantime, I will test & review the code.

bhunut-adobe:
I just added the documentation.

adorton-adobe:
@bhunut-adobe Will you please rebase this against v2 and resolve the merge conflicts?  Be sure to force push your rebased branch back your remote fork (`git push --force`) instead of merging with it (which the client will suggest that you do with a git pull).

user_sync/rules.py
```diff
@@ -799,6 +809,13 @@ def update_umapi_users_for_connector(self, umapi_info, umapi_connector):
         umapi_info.set_umapi_users_loaded()
         return user_to_group_map
 
+    def get_umapi_user_in_groups(self,umapi_info, umapi_connector, groups):
```
adorton-adobe: You can remove `self` from the argument list and make this a static method. (add the `@staticmethod` decorator)

user_sync/rules.py
```diff
@@ -743,14 +746,21 @@ def update_umapi_users_for_connector(self, umapi_info, umapi_connector):
         if self.will_process_strays:
             self.add_stray(umapi_info.get_name(), None)
 
+        if self.options['adobe_group_filter']:
```
adorton-adobe: We're probably safer changing this to `if self.options.get('adobe_group_filter') is not None`.

user_sync/rules.py
```diff
@@ -743,14 +746,21 @@ def update_umapi_users_for_connector(self, umapi_info, umapi_connector):
...
         # Walk all the adobe users, getting their group data, matching them with directory users,
         # and adjusting their attribute and group data accordingly.
-        for umapi_user in umapi_connector.iter_users():
+        for umapi_user in umapi_users:
             # get the basic data about this user; initialize change markers to "no change"
             user_key = self.get_umapi_user_key(umapi_user)
             if not user_key:
                 self.logger.warning("Ignoring umapi user with empty user key: %s", umapi_user)
                 continue
+            if umapi_info.get_umapi_user(user_key):
```
adorton-adobe: Explicit None check here as well.

user_sync/config.py
```diff
@@ -519,6 +543,10 @@ def get_rule_options(self):
         if options.get('directory_group_mapped'):
             options['directory_group_filter'] = set(six.iterkeys(self.directory_groups))
 
+        # set the adobe group filter from the mapping, if requested.
+        if options.get('adobe_group_mapped'):
```
adorton-adobe: I would pass False as the second argument to get() to make it default to False if the key is not defined.

adorton-adobe:
@bhunut-adobe I'd like to get this wrapped up soon so we can include it in the 2.4 release.  Let me know if you have any questions about the comments I've made.  It also still needs to be rebased against v2 so we can make sure these changes will work with those that have been pushed to v2 since this PR was created.  Let me know if you have any questions on the rebase process.

user_sync/config.py
```diff
@@ -247,6 +248,29 @@ def load_invocation_options(self):
                                              (username_filter_pattern, e))
                 options['username_filter_regex'] = compiled_expression
 
+        # --adobe-users
+        if self.args['adobe_users']:
```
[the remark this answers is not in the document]
adorton-adobe: just checked and it will be initialized to None

Issue #365, opened 2018-06-21
using adobe-only-user-list does not work


Luci2015:
tried it on version 2.3rc4.
I think it stopped working since the invocation_defaults got added.
The invocation_defaults also provides this option, but it keeps the 'users' argument as well, so it conflicts.

In v2.2.2 this works.

adorton-adobe:
Thank you for the report.  I will take a look.

adorton-adobe:
Fixed by #370

PR #370, opened 2018-07-26
Simple solution for #365

janssenda-adobe:
Ignore the users option if set to default

### DIFF
--- a/user_sync/config.py
+++ b/user_sync/config.py
@@ -90,6 +90,8 @@ class ConfigLoader(object):
         """Merge the invocation option defaults with overrides from the main config and the command line.
         :rtype: dict
         """
+        # copy instead of direct assignment to preserve original invocation_defaults object
+        # otherwise, setting options also sets invocation_defaults (same memory ref)
         options = self.invocation_defaults.copy()
 
         # get overrides from the main config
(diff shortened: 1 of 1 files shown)


adorton-adobe:
I still get an error:

```
λ  py .\dist\user-sync.pex -c .\config\user-sync-config.yml --process-groups --users mapped -t --adobe-only-user-action remove --adobe-only-user-list .\adobe_only.csv
2018-07-27 18:48:52 15500 INFO config - Using main config file: .\config\user-sync-config.yml (encoding utf8)
2018-07-27 18:48:52 15500 CRITICAL main - You cannot specify both a --users arg and an --adobe-only-user-list arg
```

janssenda-adobe:
I believe that is as intended, since --users was specified on the command line.  We could bypass any command line arg instead if you would prefer

user_sync/config.py
```diff
@@ -186,10 +188,14 @@ def load_invocation_options(self):
...
-        elif options['users'] and options['adobe_only_user_list']:
-            raise AssertionException('You cannot configure both a default "users" option (%s) '
-                                     'and a default "adobe-only-user-list" option (%s)' %
-                                     (' '.join(options['users']), options['adobe_only_user_list']))
+
+        # Check both that options['users'] exists, AND that it is not the default value before checking against
+        # the adobe_only_user_list parameter -  otherwise, defaults cause exception to be thrown unconditionally when
+        # the user list option is specified
+        elif (options['users'] and options['users'] != self.invocation_defaults['users']) and options['adobe_only_user_list']:
+                raise AssertionException('You cannot configure both a default "users" option (%s) '
```
adorton-adobe: @janssenda-adobe Will you please fix the indentation on this raise statement so it's just one level indented from the elif block?  It doesn't seem to be causing any issues, but a linter would probably complain about it, and we want the diff to render accurately.

janssenda-adobe: @adorton-adobe Thanks! Fixed that

PR #373, opened 2018-08-01
Logic correction to check for adobe_only_users (issue #375)

janssenda-adobe:
In response to #375 

When reading from user-sync-config.yml, the adobe_only_users key was superceeded by options['users'].  Swapped comparison to take the adobe_only_users key instead if present.  

Refactored and updated logical comparison to account for both invocation defaults AND command line options when checking for the users + adobe_only_user_list parameters so that any combination of the two is disallowed.  Previously, it was possible to specify --users on the command line even if the adobe_only_user_list was specified in the configuration file (it was ignored then).  Now, the appropriate exception is thrown.


adorton-adobe:
Thanks - I'll take a look this morning.  In the meantime, will you please create an issue for this so we can track it?  Then add a "Fixes #xxx" to your PR summary?

PR #378, opened 2018-08-06
Bug fix: --user-filter and config invocation default precedence

bhunut-adobe:
When specify --user-filter manually via cmd line. The value does not set in console output.

When specifying --user-filter and invocation default. --user-filter does not override with arg value instead it listed config file value.

#379


janssenda-adobe:
Can we get the same behavior using the following?

username_filter_pattern = options['user_filter'] = self.args['user_filter'] or options['user_filter']

adorton-adobe:
@janssenda-adobe I like the idea, but `self.args['user_filter']` will evaluate to False if it set to certain non-None values such as 0, [] or "" (empty string).  I think we're safer going with the explicit check that it isn't None.  We could do a one-liner in the form `Y if X else Z` but I think it's clearer to leave it as it is.

Issue #381, opened 2018-08-13
Invocation Defaults doesn't work for "--users file"


janssenda-adobe:
When trying to use the users key in user-sync-config.yml in the following way:

For argument --users, the default is 'all'.
users: file users-file.csv

The following error is encountered:

2018-08-13 09:02:48 18248 CRITICAL main - Unknown option "file users-file.csv" for users

The same string works on the command line.

Luci2015:
Is the connector value in invocation_defaults **csv** or still **ldap**?
Have you removed the 'all' or 'mapped' value from users in invocation_defaults?

Luci2015:
I don't think you need to do this change as it was not intended to be run along users attribute, but connector one.

Here is how I run with command line:
2018-08-22 02:48:23 2824 INFO main - ------- Command line arguments -------
2018-08-22 02:48:23 2824 INFO main - --users file users-file.csv -t
2018-08-22 02:48:23 2824 DEBUG main - -------- Resulting invocation options --------
2018-08-22 02:48:23 2824 DEBUG main -   adobe_only_user_action: ['remove']
2018-08-22 02:48:23 2824 DEBUG main -   adobe_only_user_list: None
2018-08-22 02:48:23 2824 DEBUG main -   connector: ['ldap']
2018-08-22 02:48:23 2824 DEBUG main -   process_groups: True
2018-08-22 02:48:23 2824 DEBUG main -   strategy: sync
2018-08-22 02:48:23 2824 DEBUG main -   test_mode: True
2018-08-22 02:48:23 2824 DEBUG main -   update_user_info: False
2018-08-22 02:48:23 2824 DEBUG main -   user_filter: None
2018-08-22 02:48:23 2824 DEBUG main -   users: ['all']
2018-08-22 02:48:23 2824 DEBUG main -   directory_connector_type: csv
2018-08-22 02:48:23 2824 DEBUG main -   remove_strays: True
2018-08-22 02:48:23 2824 DEBUG main -   directory_connector_overridden_options: {'file_path': 'users-file.csv'}

Here is how I run with invocation_defaults only:
2018-08-22 02:49:19 3740 INFO main - ------- Command line arguments -------
2018-08-22 02:49:19 3740 INFO main - -t
2018-08-22 02:49:19 3740 DEBUG main - -------- Resulting invocation options --------
2018-08-22 02:49:19 3740 DEBUG main -   adobe_only_user_action: ['remove']
[...]

PR #382, opened 2018-08-13
Fix for issue #381 / Unparsed config args

janssenda-adobe:
https://github.com/adobe-apiplatform/user-sync.py/issues/381

Invocation Defaults for user-sync-config now parsed if they are list type correctly.  Added small parser for ID to do this.  It parses the key values such that they match the corresponding command line values.

For example, specifying groups in the invocation defaults now yields an input list of:
['group', 'GroupA,GroupB']

Previously, it was parsed as a single value like this:
['group "GroupA,GroupB"]

Which could not be intepreted as a valid argument.

This fix resolves other list inputs as well:
eg, adobe-only-user-action write-file xyz.csv now works too from invocation defaults.

### DIFF
--- a/user_sync/config.py
+++ b/user_sync/config.py
@@ -105,7 +105,7 @@ class ConfigLoader(object):
                 elif isinstance(v, list):
                     val = invocation_config.get_list(k, True)
                     if val:
-                        options[k] = val
+                        options[k] = self.parse_args(val[0])
                 else:
                     val = invocation_config.get_string(k, True)
                     if val:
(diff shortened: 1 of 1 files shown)


adorton-adobe:
@janssenda-adobe - It's not clearly documented, but I believe the intent is for compound arguments such as `--users group "GroupA,GroupB"` or `--users file users.csv` to be specified as arrays.

`users: ['group', 'GroupA,GroupB']`
`users: ['file', 'users.csv']`

If a string is passed to list-type invocation default (like connector or users), it will be added to a single-element list in `DictConfig.get_list()`.

```python
        if value is not None and not isinstance(value, list):
            value = [value]
```

Thank you for looking into this, but I'm going to close this PR since it's a documentation issue and not a bug with the config module.  I've already created an issue to update the documentation - #387 

Would you mind reviewing the doc update when I finish it?

janssenda-adobe:
@adorton-adobe absolutely!

Issue #323, opened 2018-01-04
MemberOf or IsMemberOf is not available on LDAP server


bhunut-adobe:
We have a couple of customers whose LDAP directory does not have MemberOf or IsMemberOf or similar virtual attribute enabled or supported on the LDAP server.

For UST to work for these customers, We will need to do 2 steps lookup.

We will need to look for Group object for group membership then take group membership DN and do a second lookup for Person object.

I believe pre UST 2.0 didn't utilize memberOf attribute. Is there a way to do this for the current version?

bhunut-adobe:
@adobeDan  I added a feature to UST to support two-steps lookup. What do you think?
https://github.com/bhunut-adobe/user-sync.py/commit/3520590294a66a80c9c4050e19701d080ed884b2

adobeDan:
Yes, this is worth doing for those few people who don't have the memberOf extension loaded.  Look at the comments I left on your commit and make those changes before you create a PR.  Also you'll need to update the actual docs; work with Phil on that.

One of the issues here is that some systems don't put the dn in the attribute, they put the oid (especially for uniqueMemberOf).  We don't collect or keep OIDs anymore, so there's no way to look up by that.  You might want to think about whether there's an easy way to go from the OID to the DN.

phil-levy:
For documentation, you can add a section to the chapter on random topics (in the User Manual, section Advanced Configuration).  Just add a section at the end called "Special Configuration If IsMemberOf Is Not Available on Your LDAP Server".  The file is 
https://github.com/adobe-apiplatform/user-sync.py/blob/v2/docs/en/user-manual/advanced_configuration.md
(but in your fork).

adobeDan:
@bhunut-adobe I'd like to get 2.3 out sometime in the next week or two.  Do you think you'll have a chance to get this done?  If not it will have to wait for 2.3.1.

tobiasbp:
@adobeDan : Any news on a 2.3 release with this feature added?

thong1github:
Is this feature still under development, or will it remain unsupported?  @adobeDan @bhunut-adobe

PR #389, opened 2018-08-30
Two Steps Lookup

bhunut-adobe:
This allow direct group member lookup without utilizing isMemberOf or memberOf virtual attribute.

Nested Group Supported.

Feature for #323


user_sync/connector/directory_ldap.py
```diff
@@ -232,13 +232,15 @@ def find_group_dn(self, group):
...
 
-    def iter_group_member_dns(self, group_dn, member_attribute, searched_dns = []):
+    def iter_group_member_dns(self, group_dn, member_attribute, searched_dns = None):
         """
         return group memberships dns from specified membership attribute in LDAP group object
         :type group: str
         :type member_attribute: str
         :rtype iterable(str)
         """
+        if searched_dns == None:
```
bajwa-adobe: Would recommend check to be changed to `if searched_dns is None`.  See [best practice](https://docs.quantifiedcode.com/python-anti-patterns/readability/comparison_to_none.html).

user_sync/connector/directory_ldap.py
```diff
@@ -86,6 +86,22 @@ def __init__(self, caller_options):
         username = builder.require_string_value('username')
         builder.require_string_value('base_dn')
         options = builder.get_options()
+
+        if options['two_steps_lookup']:
```
adorton-adobe: This doesn't seem to be requiring `group_member_attribute_name` if the two-step option is enabled.  For example, sync runs if I explicitly set `two_steps_lookup` to an empty dictionary object.

```yaml
two_steps_lookup: {}
```

Checking that `options['two_steps_lookup'] is not None` should validate that an empty object isn't getting set.

user_sync/connector/directory_ldap.py
```diff
@@ -86,6 +86,22 @@ def __init__(self, caller_options):
...
+            ts_config = caller_config.get_dict_config('two_steps_lookup', True)
+            ts_builder = user_sync.config.OptionsBuilder(ts_config)
+            ts_builder.require_string_value('group_member_attribute_name')
+            ts_builder.set_bool_value('nested_group', False)
+            self.two_steps_lookup = True
+            options['two_steps_lookup'] = ts_options = ts_builder.get_options()
+            if options['group_member_filter_format']:
+                raise AssertionException(
+                    'Cannot define both group_member_attribute_name and group_member_filter_format in config')
```
adorton-adobe: Can we put the config option names in single quotes to make them stand out?

user_sync/connector/directory_ldap.py
```diff
@@ -377,7 +438,33 @@ def format_ldap_query_string(query, **kwargs):
             escaped_args[k] = six.text_type('').join(escaped_list)
         return query.format(**escaped_args)
 
+    def format_group_user_filter (self, group_dn):
```
adorton-adobe: There shouldn't be a space between the function name and the open paren.

user_sync/connector/directory_ldap.py
```diff
@@ -130,51 +146,70 @@ def load_users_and_groups(self, groups, extended_attributes, all_users):
...
-                for user_dn, user in self.iter_users(group_user_filter, extended_attributes):
-                    user['groups'].append(group)
-                    group_users += 1
-                self.logger.debug('Count of users in group "%s": %d', group, group_users)
+                if self.two_steps_lookup:
+                    for user_dn in self.iter_group_member_dns(group_dn, group_member_attribute_name):
+                        #check to make sure user_dn are within the base_dn scope
+                        if self.is_dn_within_base_dn_scope(base_dn, user_dn):
+                            #replace base_dn with user_dn and filter with all_users_filter to do user lookup based on DN
+                            result =  list(self.iter_users(user_dn, all_users_filter, extended_attributes))
```
adorton-adobe: PEP-8 nitpicks: Too many spaces after `=` operator, and there should be a space after the `#` on the two preceding comments.

user_sync/connector/directory_ldap.py
```diff
@@ -377,7 +438,33 @@ def format_ldap_query_string(query, **kwargs):
...
+        if not group_member_subfilter.startswith('('):
+            group_member_subfilter = six.text_type('(') + group_member_subfilter + six.text_type(')')
+        user_subfilter = self.options['all_users_filter']
+        if not user_subfilter.startswith('('):
+            user_subfilter = six.text_type('(') + user_subfilter + six.text_type(')')
+        group_user_filter = six.text_type('(&') + group_member_subfilter + user_subfilter + six.text_type(')')
+        return group_user_filter
 
+    @staticmethod
+    def is_dn_within_base_dn_scope(base_dn, dn):
```
adorton-adobe: We should consider using a proper DN parser instead of trying to do it ourselves.  `python-ldap` should have one, but I don't know how easy it will be to use.  If it's reasonable to do so, I would love to see that used here before we merge the PR.  If it's not feasible, I'm OK waiting until the ldap3 switch because ldap3 [also has a DN parser](https://ldap3.readthedocs.io/ldap3.utils.dn.html).

bhunut-adobe:
Will submit a new PR with new changes and rebase

PR #397, opened 2018-09-14
Allow max_adobe_users value in percentage format

bhunut-adobe:
Allows max_adobe_users to be in percentage format to allow it to be more dynamic.

#398


user_sync/config.py
```diff
@@ -497,7 +497,19 @@ def get_rule_options(self):
...
+            if percent_pattern.match(max_missing):
+                max_missing_percent = float(max_missing.strip('%'))
+                if max_missing_percent >= 0.0 and max_missing_percent <= 100.0:
+                    options['max_adobe_only_users'] = max_missing
+                else:
+                    raise AssertionException("max_adobe_only_users value must be less or equal than 100%")
+            else:
+                raise AssertionException("Not a valid value for max_adobe_only_users. Must be a percentage or fix integer")
+        else:
+            options['max_adobe_only_users'] = max_missing
```
adorton-adobe: I get an error when I set `max_adobe_only_users` to an integer as a string.  e.g.

```yaml
  max_adobe_only_users: "250"
```

We need to be able to specify the number in this manner to preserve backward compatibility.  I think you could also simplify your validation steps:

1. If `max_adobe_only_users` is a string and contains a "%" symbol, validate with the regex and check percentage bounds.
2. Otherwise cast the value `max_adobe_only_users` to an int and if you catch a ValueError, raise an ArgumentException that the value must be an integer.  note: you may need to wrap the ValueError check in a helper function that returns a bool.

```python
try:
  int(max_missing)
except ValueError:
  # raise or return from helper func
```

user_sync/rules.py
```diff
@@ -502,6 +502,9 @@ def process_strays(self, umapi_connectors):
             self.write_stray_key_map()
         if self.will_manage_strays:
             max_missing = self.options['max_adobe_only_users']
+            if isinstance(max_missing, str) and '%' in max_missing:
+                percent = float(max_missing.strip('%'))/100
+                max_missing = int((self.primary_user_count - self.excluded_user_count) * percent)
```
adorton-adobe: Since you're already parsing the percentage as an integer in the config (in `max_missing_percent`), can't you save that to the `options` dict and use that here?  If the `max_missing_percent` key is in `options` and doesn't contain a null value, use it to derive `max_missing`, otherwise use the value of `max_adobe_only_users`.  You can even convert it to a float in the config.

user_sync/config.py
```diff
@@ -497,7 +497,19 @@ def get_rule_options(self):
 
         # get the limits
         limits_config = self.main_config.get_dict_config('limits')
-        options['max_adobe_only_users'] = limits_config.get_int('max_adobe_only_users')
+        max_missing = limits_config.get_value('max_adobe_only_users',(int, str),False)
+        if isinstance(max_missing, str):
+            percent_pattern = re.compile("(\d*(\.\d+)?%)")
+            if percent_pattern.match(max_missing):
+                max_missing_percent = float(max_missing.strip('%'))
+                if max_missing_percent >= 0.0 and max_missing_percent <= 100.0:
```
adorton-adobe: You can simplify this comparison:

```python
if 0.0 <= max_missing_percent <= 100.0:
```

docs/en/user-manual/configuring_user_sync_tool.md
```diff
@@ -198,9 +198,9 @@ directory groups and Adobe product configurations and user
 groups.
     - **directory_users** can also contain keys that set the default country
 code and identity type.  See the example configuration files for details.
-- The **limits** section sets the `max_adobe_only_users` value that 
-prevents User Sync from updating or deleting Adobe user accounts if 
-there are more than the specified value of accounts that appear in 
+- The **limits** section sets the `max_adobe_only_users` value that
+prevents User Sync from updating or deleting Adobe user accounts if
+there are more than the specified value of accounts in  that appear in
```
adorton-adobe: This extra "in" appears to be a typo.

examples/config files - basic/1 user-sync-config.yml
```diff
@@ -201,6 +201,7 @@ limits:
   # job from doing something in these situations that is hard to repair, such as
   # removing or deleting a large number of users or entitlements, you can specify
   # the maximum number of Adobe-only users your User Sync job is expected to find.
+  # this number can be a fix number eg. 200 or in percentage format eg. 75%
```
adorton-adobe: Chage "a fix number" to "an integer".

user_sync/config.py
```diff
@@ -497,7 +497,19 @@ def get_rule_options(self):
...
+                    options['max_adobe_only_users'] = max_missing
+                else:
+                    raise AssertionException("max_adobe_only_users value must be less or equal than 100%")
+            else:
+                raise AssertionException("Not a valid value for max_adobe_only_users. Must be a percentage or fix integer")
```
adorton-adobe: We don't need the word "fix" here - just change it to read "a percentage or an integer".

user_sync/rules.py
```diff
@@ -502,6 +502,9 @@ def process_strays(self, umapi_connectors):
...
             if stray_count > max_missing:
                 self.logger.critical('Unable to process Adobe-only users, as their count (%s) is larger '
                                      'than the max_adobe_only_users setting (%d)', stray_count, max_missing)
```
adorton-adobe: If a percentage is in use, can put that value in the parenthesis instead of the value of `max_missing`?  

Example:

```
Unable to process Adobe-only users, as their count (5) is larger than the max_adobe_only_users setting (10%)
```

PR #414, opened 2018-10-04
Support email-type federated users with different email addresses

### LINKED ISSUE #385: Create and sync users with different usernames and email addresses
The sync tool should sync email-based federated users that have different email addresses and usernames.  It is possible to create such users in the UMAPI by first creating a user with the same email address & username and then updating the user record to have a different email address and/or username.

A common use case is if a customer's IDP sends a different primary identifier (such as userPrincipalName) for federated logins than the email address that users might use for login and collaboration.  This update would allow a separate email attribute to be specified that would map a user's email address separately from the username.

This feature doesn't introduce any new config options.  It just changes some behavior in the sync tool when both `user_email_format` and `user_username_format` are specified in `connector-ldap.yml` and map to attributes that contain different values for a given user.

Existing behavior:

If `user_email_format` and `user_username_format` are mapped to different attributes (that may contain different values from one an other), the sync tool will attempt to create a new user with the username and email address provided by the identity source.  This attempt will fail because the [create method of UserAction](https://github.com/adobe-apiplatform/umapi-client.py/blob/master/umapi_client/functional.py#L166) will not allow such a user to be created.

New behavior:

If the email values mapped from `user_email_format` and `user_username_format` differ, then the commands generated to create the user are modified so that the `create` command has the same email address as the username in the API action.  The email address will be updated in an `update` command that is appended to the user's command list before groups are assigned.

The value mapped from `user_username_format` will be the user's primary identifier.

umapi-client will also need to be updated to support this feature - https://github.com/adobe-apiplatform/umapi-client.py/issues/65

Tasks: 

- [x]  Test full sync workflow
  - [x]  User Creation
  - [x]  Group membership update
  - [x]  "preserve" adobe-only action
  - [x]  "remove" adobe-only action
  - [x]  "delete" adobe-only action
  - [x]  Update user info
  - [x]  Email address updates
- [x]  Test email address on trusted domain (especially email address update)
- [x]  Update Documentation

Note - we'll worry about testing for Okta and CSV in a future update.

adorton-adobe:
This feature adds support for email-type users that have different a different email address and username.  Such users can be created in the UMAPI by first creating the user with a consistent username and email and subsequently updating the user to specify a different email address.  Updates, removals, and deletions must use the email address (and not the username).

This doesn't fully address #385 because there is a known issue - the UMAPI returns an error when this type of user is created, removed from the org (but not deleted) and added back to the org.  Until this issue is resolved, we'll document it as a known issue in the next release candidate for 2.4.

Update - I made some changes to how these users are created and updated and the "add after remove" issue is resolved.

Fixes #385

### DIFF
--- a/docs/en/user-manual/advanced_configuration.md
+++ b/docs/en/user-manual/advanced_configuration.md
@@ -450,6 +450,37 @@ For domains that use username-based login, the `user_username_format` configurat
 
 If you are using username-based login, you must still provide a unique email address for every user, and that email address must be in a domain that the organization has claimed and owns. User Sync will not add a user to the Adobe organization without an email address.
 
+## Syncing Email-based Users with Different Email Address
+
+Some organizations must authenticate users with an internal-facing
+email-type ID such as user principal name, but wish to allow users to
+user their public-facing email address to log into Adobe products and
+use collaboration features.
+
+Internally, the Adobe Admin Console maintains a distinction between a
+user's email-type username and their email address.  These fields are
+normally set to the same value, but the Admin Console allows the
+email address to differ from the username.  The User Management API
+also supports the creation, update, and deletion of users that have
+different usernames and email addresses.
+
+**Note:** Any domain used in the email address field **must** be
+claimed and added to an Adobe identity directory.
+
+To use this functionality in the Sync Tool, simply specify both the
+`user_email_format` and the `user_username_format` options in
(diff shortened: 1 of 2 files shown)


bhunut-adobe:
Just a friendly status update. I'm reviewing it now

adorton-adobe:
@bhunut-adobe will you have time to review this some time this week?

PR #421, opened 2018-10-31
Avoid App Termination when UMAPI Create Action error

bhunut-adobe:
This will skip over the problematic users instead terminate the entire sync process.

Before:

> 2018-10-31 10:04:07 78637 CRITICAL main - Error creating umapi Action: 'Hschowãlter@adobeccetest.com': Illegal email format (must be ascii, unquoted, with no comment part)

After:

> 2018-10-31 10:06:20 78692 ERROR umapi.action - Skipping user - Error creating umapi Action: 'Hschowãlter@adobeccetest.com': Illegal email format (must be ascii, unquoted, with no comment part)

#422

### DIFF
--- a/user_sync/connector/umapi.py
+++ b/user_sync/connector/umapi.py
@@ -183,7 +183,7 @@ class UmapiConnector(object):
         if len(commands) > 0:
             action_manager = self.get_action_manager()
             action = action_manager.create_action(commands)
-            if action:
+            if action is not None:
                 action_manager.add_action(action, callback)
 
 
@@ -320,7 +320,7 @@ class ActionManager(object):
                                              requestID=self.get_next_request_id())
         except ValueError as e:
             self.logger.error("Skipping user - Error creating umapi Action: %s" % e)
-            return
+            return None
         for command in commands.do_list:
             command_name, command_param = command
             command_function = getattr(action, command_name)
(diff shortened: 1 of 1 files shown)


user_sync/connector/umapi.py
```diff
@@ -318,7 +319,8 @@ def create_action(self, commands):
             action = umapi_client.UserAction(umapi_identity_type, email, username, domain,
                                              requestID=self.get_next_request_id())
         except ValueError as e:
-            raise AssertionException("Error creating umapi Action: %s" % e)
+            self.logger.error("Skipping user - Error creating umapi Action: %s" % e)
+            return
```
adobeDan: You should explicitly return `None` from value-producing functions, as it helps with type analysis and automatic error checking.  Especially when you are returning class objects otherwise, as in this case you are using `None` specifically as a flag value.
```suggestion
            return None
```

user_sync/connector/umapi.py
```diff
@@ -183,7 +183,8 @@ def send_commands(self, commands, callback=None):
         if len(commands) > 0:
             action_manager = self.get_action_manager()
             action = action_manager.create_action(commands)
-            action_manager.add_action(action, callback)
+            if action:
```
adobeDan: if there is a flag value returned on error, check for it explicitly.  Otherwise you might get surprised if someone defines `__bool__` (py2) or `__nonzero__` (py3) or `__len__` (both) on the class.
```suggestion
            if action is not None:
```

adorton-adobe: Agreed - we should be explicit with None checks whenever possible.

adorton-adobe:
@adobeDan Do you plan to review this again?  I think it's ready to merge.

Issue #416, opened 2018-10-11
Log a meaningful error if Okta connector isn't invoked correctly


adorton-adobe:
If the Okta config file is enabled and the `--connector` CLI param or invocation default is not specified, log an error that this option is required.  Right now the error message produced by the UST mentions that the "host" key can't be found in the LDAP config - this is inaccurate and misleading.

adorton-adobe:
@shasibhusanJena - I see that I assigned this to you, but I don't remember if we've discussed it.  Do you mind handling it?  Should be a pretty easy fix.  Let me know if you're able to and if you have any questions.

shasibhusanJena:
yesterday I was trying to find the line or places those have to make changes on the code, and formatting new string/ msg for the log. we can discuss and commit changes.

shasibhusanJena:
I could reproduce this error on my local, as per my understanding when I comment out 
# ldap: "connector-ldap.yml"   or  # okta: "connector-okta.yml"  
we are getting a common  error like  "value not found for key: host in LDAP configuration" 

so we could have different error stmt for different error types like 

for # ldap: "connector-ldap.yml"
"Do enable LDAP configuration to fetch key and values in the user-sync-config.yml file"

for # okta: "connector-okta.yml"
"Do enable Okta configuration to fetch key and values in the user-sync-config.yml file"

@adorton-adobe plzsuggest me if any better way of representation of log

shasibhusanJena:
it look like this is the only line will get updated.

https://github.com/adobe-apiplatform/user-sync.py/blob/2112484a84256de57e3c6e088b847289f83fef81/user_sync/config.py#L742

adorton-adobe:
The condition to get the error is pretty specific - 

1. Comment out the LDAP config
2. Enable the Okta config
3. `--connector` command line option is not specified
4. `connector` invocation option is set to "ldap" or is blank

To detect the condition in the code and report the error, I would check that the right connector config is enabled for the value of the "connector_type" option.  In this case, it would still be set to LDAP (the default).  So the error message would be something like "LDAP config must be enabled for the connector type 'ldap'"

adorton-adobe:
You don't want to modify that line.  It's a generic check to ensure that required keys are present in the config.  You need to raise the error prior to the config processor getting there.

shasibhusanJena:
below are my observations 

[...]

PR #438, opened 2019-01-17
Log a meaningful error if Okta connector isn't invoked correctly

shasibhusanJena:
added the log in app.py file, please review it. issue no #416


shasibhusanJena:
- removed line from app.py  default AssertionException block
- added a separate code block on config.py file to validate this condition.

shasibhusanJena:
raised a new pull request from Branch

PR #442, opened 2019-01-28
Log a meaningful error if Okta connector isn't invoked correctly #416 - V2 for release 2.4   

shasibhusanJena:
- removed line from app.py default AssertionException block 
-  added separate code block on config.py file to validate this condition under get_directory_connector_options definition.


shasibhusanJena:
connectors_config.value provides the option to do one round of comparison to check then 3 times in a loop.

adorton-adobe:
@shasibhusanJena why did you close this?  Do you need to make additional changes before review?

PR #443, opened 2019-01-28
V2 for release 2.4

shasibhusanJena:
- removed line from app.py default AssertionException block - added a…  …
… separate code block on config.py file to validate this condition.
-- doing a comparison with connectors_config. the value so that, don't have to loop through to make a conditional check "Okta" type

### DIFF
--- a/user_sync/config.py
+++ b/user_sync/config.py
@@ -324,9 +324,8 @@ class ConfigLoader(object):
         options = {}
         connectors_config = self.get_directory_connector_configs()
 
-        if ( 'okta' in connectors_config.value) and (connector_name == 'ldap'):
-            if connectors_config.accessed_keys is not None:
-                raise AssertionException("Failed to match request : for Okta Connector type receiving LDAP as input type from commnand line")
+        if connector_name != 'csv' and connector_name not in connectors_config.value:
+            raise AssertionException("Config file must be specified for connector type :: '{}'".format(connector_name))
 
         if connectors_config is not None:
             connector_item = connectors_config.get_list(connector_name, True)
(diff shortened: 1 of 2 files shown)


adorton-adobe:
@shasibhusanJena You don't need to make new PR when you make new commits.  Just commit to the branch the PR is coming from and they'll be picked up here.

shasibhusanJena:
oh, I was not knowing about it before, that will be easy then. thanks

user_sync/config.py
```diff
@@ -323,6 +323,11 @@ def get_directory_connector_options(self, connector_name):
         """
         options = {}
         connectors_config = self.get_directory_connector_configs()
+
+        if ( 'okta' in connectors_config.value) and (connector_name == 'ldap'):
+            if connectors_config.accessed_keys is not None:
+                raise AssertionException("Failed to match request : for Okta Connector type receiving LDAP as input type from commnand line")
```
adorton-adobe: If you make the check more generic you'll need to make this message generic as well.  I'd do something like this:

```python
raise AssertionException("Config file must be specified for connector type '{}'".format(connector_name))
```

user_sync/config.py
```diff
@@ -323,6 +323,11 @@ def get_directory_connector_options(self, connector_name):
         """
         options = {}
         connectors_config = self.get_directory_connector_configs()
+
+        if ( 'okta' in connectors_config.value) and (connector_name == 'ldap'):
+            if connectors_config.accessed_keys is not None:
```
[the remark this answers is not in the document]
shasibhusanJena: connectors_config.accessed_keys that were to find the type as csv, okta, LDAP. I was one layer of the check, which is not required now, so removing it.

user_sync/config.py
shasibhusanJena: modified. please review

Issue #465, opened 2019-03-13
Add ability to skip users from extension config


vossen-adobe:
I've had this come up a lot of times - the extension config would be greatly more useful if we had the ability to drop users from the sync conditionally.  Normally, I'd have to change a user's email to bogus values in order to prevent them from synching, but it would be great to be able to do this cleanly.

An example case:
Filtering by OU.  The LDAP syntax is not sophisticated enough to do this, but it's a breeze with regex in the extension.  The only trouble is that you still can't stop non-matching users from being pushed without the hack I mentioned above.

I can think of numerous other benefits for this as well!

adobeDan:
Why do you want to do this in the extension code, as opposed to just using the existing "exclude_users" config option (which allows for regular expressions)?  Are you trying to provide for more sophisticated logic than just an OR of regular expressions?

vossen-adobe:
@adobeDan Yes - more sophisticated logic, but also the fact that exclude_users is applied on the UMAPI side, whereas this change would allow filtering based on AD conditions/extended attributes.  The other choice, --users-filter option, seems not to handle attribute comparison and requires an exact match  according to the documentation:

--user-filter
"Limit the set of users that are examined for syncing to those matching a pattern specified with a regular expression."

adorton-adobe:
Closing since the related PR is closed.

PR #466, opened 2019-03-13
Add ability to skip users based on criteria from extension config

vossen-adobe:
Feature code for #495

https://github.com/adobe-apiplatform/user-sync.py/issues/465

Simple addition that permits skipping of users from extension config.

Resulting behavior:
When applied to a user, they are effectively removed from the sync process.  This has the same end effect as removing them from the AD group or all users filter, except that the decision to remove them has not been made until after the comparison (owing to where the extension config hooks in).

  - if they exist on the admin console side and are skipped they will be treated like adobe-only-users.  If in a target group, they will be removed from that group.  
  - using --adobe-only-user-action flag seems to have the normal effect as for other users.  Skipping a user causes them to be affected by this flag.
  - if they do not exist on the admin console and are skipped, they will not be created /  provisioned
  - In push sync mode, users will only be created if they are NOT skipped

Example extension config code adds all users in AD OU "Specific OU" to the console group "AdobeGroupForOUUsers" and skips any user not in that OU:

```python
extended_attributes:
  - distinguishedName
extended_adobe_groups:
  - AdobeGroupForOUUsers

after_mapping_hook: | 
if "Specific OU" in source_attributes.get('distinguishedName'):
  target_groups.add('AdobeGroupForOUUsers')`
else:
  skip_user = True
```

### DIFF
--- a/user_sync/rules.py
+++ b/user_sync/rules.py
@@ -405,12 +405,10 @@ class RuleProcessor(object):
                 self.log_after_mapping_hook_scope(after_call=True)
 
                 # skip user if hook specifies true
-                skip_user = self.after_mapping_hook_scope['skip_user']
-                if type(skip_user) is bool and skip_user:
-                    self.get_umapi_info(PRIMARY_UMAPI_NAME).get_desired_groups_by_user_key().pop(user_key)
+                if self.after_mapping_hook_scope['skip_user'] is True:
                     self.directory_user_by_user_key.pop(user_key)
-                    if user_key in filtered_directory_user_by_user_key:
-                        self.filtered_directory_user_by_user_key.pop(user_key)
+                    self.filtered_directory_user_by_user_key.pop(user_key)
+                    self.get_umapi_info(PRIMARY_UMAPI_NAME).get_desired_groups_by_user_key().pop(user_key)
                     continue
 
                 # copy modified attributes back to the user object
--- a/user_sync/rules.py
+++ b/user_sync/rules.py
@@ -148,6 +148,8 @@ class RuleProcessor(object):
             'logger': logger,
             # for exclusive use by hook code; persists across calls
             'hook_storage': None,
+            # allow hook to drop users from the list
+            'skip_user': False
         }
 
         # map of username to email address for users that have an email-type username that
@@ -359,7 +361,6 @@ class RuleProcessor(object):
         directory_users = directory_connector.load_users_and_groups(groups=directory_groups,
                                                                     extended_attributes=extended_attributes,
                                                                     all_users=directory_group_filter is None)
-
         for directory_user in directory_users:
             user_key = self.get_directory_user_key(directory_user)
             if not user_key:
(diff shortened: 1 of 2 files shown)


user_sync/rules.py
```diff
@@ -403,6 +404,15 @@ def read_desired_user_groups(self, mappings, directory_connector):
                 exec(options['after_mapping_hook'], self.after_mapping_hook_scope)
                 self.log_after_mapping_hook_scope(after_call=True)
 
+                # skip user if hook specifies true
+                skip_user = self.after_mapping_hook_scope['skip_user']
+                if type(skip_user) is bool and skip_user:
```
adobeDan: The "Pythonic" way to say this (which is also much more efficient) is:
`if skip_user is True:`

vossen-adobe: Thanks for that! My brain is always in Java, or I have no excuse for missing that one :)

vossen-adobe:
Fixed issues mentioned above!

adobeDan:
@adorton-adobe Issue #465 states that this code doesn't apply on push.  I suspect that's because the extension is being applied after matching with Adobe is done, rather than before.  If it were done before, then (1) it would apply on push and (2) the code to drop the user would be much simpler: we could just remove them from the dictionary on the LDAP side.  If I get a minute I will look at what's going on here, but I thought you might want to check it out.

vossen-adobe:
@adobeDan I added some behavior examples to the post above.  The observed result is that using this skip option has the same end result as though they were not pulled from the directory in the first place - which was the intended effect.  It seems to apply on push as well - users are not created if they are skipped.

adobeDan:
Hi @vossen-adobe thanks for the additional examples and descriptions.  I agree this is a useful feature and works well as you have built it.  It's just that it has gotten me thinking about the wisdom of applying the extensions *after* doing the Adobe sync rather than before, when we are doing the directory read.  For example, suppose I wanted an extension to actually change the username or the email (e.g., to do obfuscation): I can't do that now because it's applied too late.  So I am not pushing back because I think the code is wrong; I am pushing back because I think we can make a slightly deeper additional change that will both simplify this one and allow extensions to do a lot more.  @adorton-adobe your thoughts?

Luci2015:
Source exclusion is definitely useful (especially if it's based on a regex), as it comes from time to time in discussions we have with Customers.
Another way I was able to exclude from source was to tweak the all_users_filter.

For example, I have all the users I want to exclude in some AD group named GroupName, but these users are also part of some mapped groups (and don't want them to get created/entitled on Adobe side) I can use this not memberOf filter:

all_users_filter: "(&(objectClass=user)(objectCategory=person)(!(userAccountControl:1.2.840.113556.1.4.803:=2))(!(memberOf=CN=GroupName,OU=someOU,DC=some,DC=domain)))"

OUs don't have memeberOf attribute to work with, so if the filter is requested to be that broad, then checking the distinguishedName for each user is an option.

vossen-adobe:
Closing to revisit with new approach

PR #469, opened 2019-03-19
Exclude Users - Ignore Case

### LINKED ISSUE #468: Case Sensitive when Excluding Users
When excluding users in user-sync-config.yml, UST does a regex case sensitive match where it should be case insensitive.

As a workaround, I put (?i) in front of the username to ignore case.

```
  exclude_users:
    #- ".*@adobeccetest.com"
    #- "freelancer-[0-9]+.*"
    - "(?i)MRICE@adobeccetest.com"
```

james-d-elliott:
fixes #468 

I've not tested it, but it should work. I don't think it's practical to have users use the python regex inline modes/modifiers and can't think of a situation where case sensitivity would be required - the only caveat is that I don't believe you can turn case sensitivity back on inline with python.

### DIFF
--- a/user_sync/config.py
+++ b/user_sync/config.py
@@ -495,7 +495,7 @@ class ConfigLoader(object):
                 try:
                     # add "match begin" and "match end" markers to ensure complete match
                     # and compile the patterns because we will use them over and over
-                    exclude_users.append(re.compile(r'\A' + regexp + r'\Z', re.UNICODE))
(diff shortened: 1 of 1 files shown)


james-d-elliott:
One slight issue I have just realized is that technically speaking the local part of an email IS case sensitive; just it's rarely used in a way that would cause issues for this purpose.

Issue #463, opened 2019-03-12
additional_groups not working in 2.4


cosmin-balanoiu:
[ust2.4additionalGroups.zip](https://github.com/adobe-apiplatform/user-sync.py/files/2956649/ust2.4additionalGroups.zip)

I tried every combination I could think of. I cannot get it to work.
I attached my config files and the last log. I tried this in 2.4 on python 3.6.5 and 3.6.6, in test mode and live, I always get the same result.
The AD group and the Admin Console one are named ADB-one.
 
The tool works fine if I comment out additional_groups.
 
Thanks,
Cosmin

SCRICrtoole:
I am having similar issues.  The groups that match the RegEx do not sync, while those with explicit mappings do.  

I did have to single quote the expression because it did not like the "\1" in the target line.  That got it to run, but then I do not see the groups that match my expression.

Example
I changed

- source: "(.+)-ACL"
  #     target: "ACL-Grp-(\1)"

to 

- source: '(.+)-ACL'
  #     target: 'ACL-Grp-(\1)'

It's still not working, but there is definitely an issue with the examples as they are.

cosmin-balanoiu:
any progress on this?

vossen-adobe:
@cosmin-balanoiu can you give some examples of the full group names you are trying to match?

cosmin-balanoiu:
Just "ADB-one". I tried to test this on a simple group to see if the tool picks it up after a report from a customer that this is not working. You can find the regexp I used in the attached files.

C

vossen-adobe:
I get the same issues - I'll do some digging

vossen-adobe:
Alright - two things: 

1) There seems to be a bug in the code where a string is passed instead of a list when only one group comes back for a user.  The method causing this is only called when there are additional groups mapped.  Unfortunately this means that everyone in your que must be in more than one group, even if they are not part of additional groups.   This should be likely be fixed in the next release.

2) Additional groups is targeting users who have groups in addition to being part of a normally mapped group.  In your case, only users who are in the group called "nested" will be pulled in by the additional groups method.  You can use --users all in if you needed to.  Neither of these will work correctly until the next version, however.

vossen-adobe:
Near 309 in directory_ldap:

`    def get_member_groups(self, user):
        group_names = []
        groups = LDAPValueFormatter.get_attribute_value(user, 'memberOf')
        for group_dn in map(dn.str2dn, groups):
[...]

PR #472, opened 2019-04-01
#463 fix additional groups bug

vossen-adobe:
directory_ldap:
- Account for the case when a user has only 1 group.  LDAPFormatter returns a string instead of a list because first_only is set to true.  We cannot set this to false, because the other attributes depend on it.   We also cannot modify the dn method as it is not part of User Sync.  So, the best solution is to create a one item list in this case.

- If a user has no groups (--users all), LDAPFormatter returns None for the group list -- in this case, we need to immediately return instead of trying to call map() on the NoneType object.

examples\user-sync-config.yml:
- Update the regular expression to escape the "\\" character.  Tested working this way.  Also verified that configapp is able to maintain these fields.  After processing by the app, the following transformation takes place:
```
     source: "ACL-(.+)"               ->          source: ACL-(.+)
     target: "ACL-Grp-(\\1)"          ->          target: ACL-Grp-(\1)
```
And the mapping works in both cases.

### DIFF
--- a/user_sync/connector/directory_ldap.py
+++ b/user_sync/connector/directory_ldap.py
@@ -396,7 +396,7 @@ class LDAPDirectoryConnector(object):
 
         if not groups:
             return group_names
-        elif type(groups) is str:
+        elif isinstance(groups, str):
             groups = [groups]
 
         for group_dn in map(dn.str2dn, groups):
--- a/user_sync/connector/directory_ldap.py
+++ b/user_sync/connector/directory_ldap.py
@@ -393,6 +393,12 @@ class LDAPDirectoryConnector(object):
         """
         group_names = []
         groups = LDAPValueFormatter.get_attribute_value(user, 'memberOf')
+
+        if not groups:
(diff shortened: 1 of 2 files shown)


user_sync/connector/directory_ldap.py
```diff
@@ -393,6 +393,12 @@ def get_member_groups(self, user):
         """
         group_names = []
         groups = LDAPValueFormatter.get_attribute_value(user, 'memberOf')
+
+        if not groups:
+            return group_names
+        elif type(groups) is str:
```
adorton-adobe: We should generally use `instanceof` to do type checking instead of `type` since `instanceof` will account for inheritance.

Example:

```python
if instanceof(groups, str):
  groups = [groups]
```

(actually, duck-type checking is the most preferred way of doing it, but it might be difficult to distinguish between a string and array without checking the type)

vossen-adobe: Done!

Issue #474, opened 2019-04-05
PEX_ROOT environment variable not working for 2.4.2


cosmin-balanoiu:
Setting the PEX_ROOT environment variable no longer works around the MAX_PATH when using 2.4.2. There is a very long path for the cryptography dependency that exceeds the limit even with the set variable. Bumped into this on two customer environments.

adorton-adobe:
Thanks, I will take a look.  We may need to pin the dependency for now.

adorton-adobe:
Can you let me know if this is the case in python 2.7 as well as 3.6?

vossen-adobe:
This was happening for me to on windows 10.  I assumed it was a local issue becausei t was just happening on one machine,  but I guess not.  We could also try to set the long path enabled in registry with the installer for now.

adorton-adobe:
The Python installer will do that (at least in the 3.6 installer).  How is the UST installer running the Python installer?  Is there a CLI option we can pass to the Python installer to make it enable long path support?

adorton-adobe:
The more I think about it, the clearer it is that we need to pin Cryptography to an older version.  I think we should definitely have the UST installer enable long path support for Windows 10 and Server 2016.  But that won't solve the problem for Server 2012.  

cryptography is required by the umapi-client, so we'll pin the version there and release a new umapi-client, and with it a new UST release.

vossen-adobe:
I just pushed an installer update that should resolve for win 10 / 2016 using the registry, but it doesn't work for 2012

adorton-adobe:
Turns out the issue is with pycryptodome and not cryptography.  We don't need a umapi-client build, so I will plan to make a new UST release shortly.

adorton-adobe:
fixed in v2.4.3

Issue #436, opened 2019-01-09
Switch to ldap3


adorton-adobe:
python-ldap is a pain point for build management.  The team that maintains it does not provide builds for win64 and building it from source on win64 is fairly complicated.  We currently rely on an unofficial repository of win64 builds of many popular Python dependencies.  Reliance on this source isn't a good idea for long term project sustainability.  It also prevents us from supporting Python 3.7 because we're pegged to an older version of python-ldap that doesn't support 3.7.  Upgrading that dependency will require extensive testing (which we'll need to do anyway if we switch to ldap3).

[ldap3](https://ldap3.readthedocs.io/) is a mature, pure Python ldap client library that won't present any of the problems we have with python-ldap.  It additionally has a nicer, more Pythonic interface and may make it easier to support certain features that customers have requested (such as StartTLS).

We should plan to migrate the ldap connector to use ldap3 and target that update to the 2.5 release.

adorton-adobe:
This is merged so we can close this.

PR #475, opened 2019-04-05
Switch Python-LDAP to LDAP3

bhunut-adobe:
Switching to Pure Python LDAP library.

New Feature:
Additional support for Anonymous and NTLM Authentication for LDAP Binding

Solution for Issue #436


adorton-adobe:
We need to make a new release before we can merge this to fix #474.  Once that is done I'll review this.

In the meantime, can you rebase this against v2?  That should resolve the merge conflict.  Be sure to force-push the rebase back to your feature branch.  Let me know if you want to walk through process - I'm happy to help.

examples/config files - basic/connector-ldap.yml
```diff
@@ -22,6 +22,15 @@ password: "LDAP password goes here"
 host: "ldap://ldap.example.com"
 base_dn: "DC=example,DC=com"
 
+# (optional) You can specify what Authentication method to bind LDAP
+# connection with. You can choose either Anonymous, Simple or NTLM.
+# If username is not specified above, authentication method will default to anonymous
```
adorton-adobe: I think it's important to use the work "override" here since that is what the connector is doing.  So the note would be something like `If username is not specified above, the LDAP connector will override the auth method and set it to 'anonymous'`

I also think it is important to the note that this line replaced - `If you choose anonymous, you don't have to specify username and password above.`.

user_sync/connector/directory_ldap.py
```diff
@@ -81,22 +78,38 @@ def __init__(self, caller_options):
...
         self.user_country_code_formatter = LDAPValueFormatter(options['user_country_code_format'])
 
-        password = caller_config.get_credential('password', options['username'])
+        auth_method = options['authentication_method'].lower()
+
+        if options['username'] is not None:
+            password = caller_config.get_credential('password', options['username'])
+        else:
+            # override authentication method to anonymous if username is not specified
+            auth_method = 'anonymous'
```
adorton-adobe: I think it would be very helpful to users to log a message here if we are going to override the auth method.  I agree that we need this feature, but I don't want to cause confusion.  It can a simple info message like "username not specified, overriding authentication method to 'anonymous'".

bhunut-adobe: > 2019-04-09 10:22:05 98627 WARNING ldap - username not specified, overriding authentication method to 'anonymous'

Like this? Should it be warning or info?

adorton-adobe: I think INFO is fine.  We're just trying to make users aware of this documented feature.

Issue #477, opened 2019-04-10
adobe-only-user-list broken since 2.3


vossen-adobe:
Intially bug introduced in this commit: cb82a0ab513ff38b37c70830faf849f458816dc7

The option for adobe-only-user-list seems not to have worked since the additional groups feature.  When using the option, no directory is pulled into app.py.   This causes the statement on line 317 to produce a null pointer on the state attribute:

directory_connector.state.additional_group_filters = additional_group_filters

Solution is simple - the additional group filter should only be assigned when a connector exists, since it doesn't apply to the adobe-only-user-list.  The output is:

2019-04-10 14:48:42 7064 INFO main - -t -c ..\test_config\user-sync-config.yml --adobe-only-user-list ..\test_config\remove-list.csv
2019-04-10 14:48:42 7064 INFO main - -------------------------------------
2019-04-10 14:48:42 7064 ERROR main - Unhandled exception
Traceback (most recent call last):
  File "D:/Repositories/adobe/user-sync-fork/user_sync/app.py", line 85, in main
    begin_work(config_loader)
  File "D:/Repositories/adobe/user-sync-fork/user_sync/app.py", line 322, in begin_work
    directory_connector.state.additional_group_filters = additional_group_filters
AttributeError: 'NoneType' object has no attribute 'state'
2019-04-10 14:48:42 7064 INFO main - ========== End Run (User Sync version: 2.4.3) (Total time: 0:00:00)

adorton-adobe:
Can you look at #459 and tell me if that fixes this issue?  I think it may.

vossen-adobe:
Indeed - fixed by #459!

bajwa-adobe:
`adobe-only-user-list` seems still broken on my side in latest build 2.5.0rc2 but different error. And it is also failing on v2.4.3 and gives different error. 

Environment: Windows/Python 3.6.8 

I got following error on v2.5.0rc2:

```
>python user-sync.pex -t --adobe-only-user-list remove-list.csv --adobe-only-user-action remove
2019-05-14 15:04:17 21300 INFO config - Using main config file: user-sync-config.yml (encoding utf8)
2019-05-14 15:04:17 21300 INFO config - Adobe-only user list specified, ignoring 'users' setting
2019-05-14 15:04:17 21300 ERROR main - Unhandled exception
Traceback (most recent call last):
[...]

PR #478, opened 2019-04-10
#477 fix (adobe-only-user-list)

vossen-adobe:
Fix for #477

Added conditional to skip additional groups if no connector is specified.  For now, additional groups only applies to LDAP -- but I thought it might be worth skipping for no connector vs loading only for ldap, since there remains the option in the future to add to other connector types.


vossen-adobe:
Ooops - I referenced the wrong commit.  I'll redo it.

Issue #486, opened 2019-05-09
user-sync-v2.5.0rc2-win64-py368.zip


Luci2015:
Did not had time to look thoroughly on it, but I've had a test with this version today (user-sync-v2.5.0rc2-win64-py368.zip)  and it did not process the groups (and maybe missing warnings about country code).

Luci2015:
So, I've investigated a bit further and it seems that if I run the tool with the invocation defaults where I have **Yes** or **True** for  process_groups, update_user_info, test_mode
=> these are all ignored and it used False by default for all, including test_mode, which has potential to do some serious damage on run
This needs a fix asap.

adorton-adobe:
Please post your user sync config and the CLI parameters that your are using.

Luci2015:
I'll do a PR in a few minutes. Stay tuned.

Luci2015:
Can you please review this?

bajwa-adobe:
The version cmd also seems broken on new build 

`python user-sync.pex --version` results in error `RuntimeError: Could not determine version`

PR #487, opened 2019-05-10
 Fix for #486

Luci2015:
Offending code is this:
`
for k, arg_val in self.args.items():
            if arg_val is None:
                continue
            options[k] = arg_val
`
The bool values do not take None as default, they take False, so you won't know if the user forced this value or not in command line unless a default is set in click.options to meet the None requirement.

### DIFF
--- a/user_sync/app.py
+++ b/user_sync/app.py
@@ -125,7 +125,7 @@ def main():
               cls=user_sync.cli.OptionMulti,
               type=list,
               metavar='ldap|okta|csv [path-to-file.csv]')
-@click.option('--process-groups/--no-process-groups',
+@click.option('--process-groups/--no-process-groups', default=None,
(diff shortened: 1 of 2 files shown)


adorton-adobe:
I originally had defaults set for all options, but I had problems with the CLI args taking precedence over `invocation_defaults` when they should not be.

ancientHacker:
Yes, this bug came in with the introduction of `click` for command-line processing (a change I heartily agree with, by the way).  You have a couple of possible fixes: click does allow None as a default value for booleans, or you can define these values as feature flags instead of booleans.

adorton-adobe:
@ancientHacker I think you're right - None is the default when no default is specified, so this should work.  I'll do some testing and get this merged.

PR #495, opened 2019-05-28
Looks like yield is releasing double rows for py2 - this works

vossen-adobe:
Same as #493
In py2, I think the row is yielded twice.  Somehow this didn't show up until after the merge.

### DIFF
--- a/user_sync/helper.py
+++ b/user_sync/helper.py
@@ -116,7 +116,7 @@ class CSVAdapter:
                         newrow = {}
                         for key, val in six.iteritems(row):
                             newrow[key.decode(encoding, 'strict')] = val.decode(encoding, 'strict') if val else val
-                        yield newrow
+                        row = newrow
                     yield row
             except UnicodeError as e:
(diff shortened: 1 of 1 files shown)


user_sync/helper.py
```diff
@@ -116,7 +116,7 @@ def read_csv_rows(cls, file_path, recognized_column_names=None, logger=None, enc
                         newrow = {}
                         for key, val in six.iteritems(row):
                             newrow[key.decode(encoding, 'strict')] = val.decode(encoding, 'strict') if val else val
-                        yield newrow
+                        row = newrow
```
adorton-adobe: I didn't notice this before but it looks like you have two `yield` statements - one on line 119 and another on line 120.  Remove `row = newrow` because we still don't want to mutate the `row` object.  Just have it yield once on line 120 and we should be OK.

adorton-adobe: Actually, never mind - we should still yield `newrow` on line 119, but put the `yield` on 120 under and else clause so we only `yield row` when not in py2

vossen-adobe: Done!

Issue #95, opened 2017-03-18
overhaul of names and usage of config settings and command-line args


adobeDan:
The "Enterprise Dashboard" has been rebranded to the "Enterprise Admin Console" everywhere but, apparently, in User Sync.  While this would be a rather widespread change, affecting not only the code but also the configuration options and configuration file names, and while it would *not* be backward compatible, it's one that I believe we should make sooner rather than later.

Question: if we did this change, should we maybe bump the major version number?  After all, it is backwards-incompatible, and we have been talking for quite a while about the "2.0" release of user-sync.  I'm thinking, what with all the features in this 1.2 release, we should maybe rename it the 2.0 release and do the rename...

adobeDan:
@phil-levy Need you to look at this before you go on vacation.

phil-levy:
I thought it was rebranded to "Admin Console" to cover more than enterprise and as far as I know that is the term we use throughout the documentation.

"dashboard" is used in the configuration in a couple of places.  I don't think it is worth the time, expense, and disruption to make this change but if Neal wants to I have no objection.  We could make it compatible by recognizing both tags so it may just be one place.

I would not make changes in the code that are not manifested externally.

adobeDan:
Ok I will check with Neal. Good point about accepting both forms for backward compatibility. 

It's actually a very easy global replace at the code level, because they are both distinctive terms.

NealMirchAdobe:
If possible, it would be best to replace any reference of "Enterprise Dashboard" with "Admin Console".  Accepting both would be crafty!

adobeDan:
@phil-levy @NealMirchAdobe I don't think we have enough customers yet that we need to be overly concerned with backward compatibility.  As long as we can give simple, global search/replace instructions that go along with a release, I think we have few enough users that we can just put it in the release notes to update their configs/scripts and we can have AGS work with the few existing customers to help them port.

@adorton-adobe @bhunut-adobe what do you think?

adorton-adobe:
Seems reasonable to me.  I ran into this earlier today on a call.  It wasn't difficult to figure out which keys needed to be renamed in the config.

adobeDan:
Sorry, but I disagree strongly.  We already have compatibility problems with the official rebranding of the "Enterprise Dashboard" to "Admin Console".  See @NealMirchAdobe's comment from a week ago.  Pretending that didn't happen is not a solution.

jameslockman:
[...]

PR #513, opened 2019-07-29
removed unused methods from RuleProcessor.

michaelmernin:
Removed unused methods [get_new_groups, calculate_groups_to_add, calculate_groups_to_remove] from  Rules.RuleProcessor. 

These three methods only call themselves and are not used anywhere else in the codebase. 

It looks like this functionality became obsolete when the Enterprise Dashboard was rebranded into the admin console, for reference: issue #95 (https://github.com/adobe-apiplatform/user-sync.py/issues/95).  Which took place ~v1.2rc2.

### DIFF
--- a/user_sync/rules.py
+++ b/user_sync/rules.py
@@ -975,54 +975,6 @@ class RuleProcessor(object):
                 result.add(normalized_group_name)
         return result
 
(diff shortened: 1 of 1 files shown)


adorton-adobe:
Fixed by #523

Issue #518, opened 2019-08-01
Fix for sorting issue in write_stray_key_map


sakonduri:
A bug deep within the rules implementation for multiple umapi's. Python 3 isn't able to sort a list of [None, str] in write_stray_key_map method. Rules.py will need to be updated to reflect this use case.

```python
    # None sorts before strings, so sorting the keys in the map
        # puts the primary umapi first in the output, which is handy
        for umapi_name in sorted(self.stray_key_map.keys()):
            for user_key in self.get_stray_keys(umapi_name):
                id_type, username, domain = self.parse_user_key(user_key)
                umapi = umapi_name if umapi_name else ""
                if secondary_count:
                    row_dict = {'type': id_type, 'username': username, 'domain': domain, 'umapi': umapi}
                else:
                    row_dict = {'type': id_type, 'username': username, 'domain': domain}
                rows.append(row_dict)
```

PR #519, opened 2019-08-01
Fix for sorting issue in write_stray_key_map #518

sakonduri:
This fix is for  (https://github.com/adobe-apiplatform/user-sync.py/issues/541)

The original code is written:

```python
        # None sorts before strings, so sorting the keys in the map
        # puts the primary umapi first in the output, which is handy
        for umapi_name in sorted(self.stray_key_map.keys()):
            for user_key in self.get_stray_keys(umapi_name):
                id_type, username, domain = self.parse_user_key(user_key)
                umapi = umapi_name if umapi_name else ""
                if secondary_count:
                    row_dict = {'type': id_type, 'username': username, 'domain': domain, 'umapi': umapi}
                else:
                    row_dict = {'type': id_type, 'username': username, 'domain': domain}
                rows.append(row_dict)
```

In the PR, we simply remove the sort so that comparisons between None and str are no longer made.  Users can sort users as they wish in the CSV.  The umapi column makes it clear which users belong to which umapi.


adorton-adobe:
What problem does this solve?  I don't understand the issue you're referencing.

Does this relate to an open issue?  If so, please edit your PR description and put this as the last line:

fixes #XXX

Where XXX is the issue number.

If there isn't a related issue, please create one and describe the bug, the impact it has, steps to reproduce, and recommended fix.  (and add the "fixes" statement to this PR)

vossen-adobe:
@sakonduri reminder - can you update this?

vossen-adobe:
https://github.com/adobe-apiplatform/user-sync.py/issues/541

Updated fix

adorton-adobe:
How do I replicate this?  I just set up a test where I have 5 users in a "primary" console and 5 in a "secondary" console.  Both sets of users belong to the same directory, but each has unique email address (there is no overlap between the two sets).  When I run the tool with `--adobe-only-user-action write [filename]`, I don't get an error.  The file is written, but only with the 5 users I added to the primary console.

I'll dig into it more and see if I've misconfigured something.

This also raises the question - do we even need to sort the output here?  It doesn't make any functional difference as far as I can tell.

(I'm testing this in Python 3.6)

vossen-adobe:
I've just reproduced by making sure there are adobe-only users on both sides by using '--user-filter x'.  I have a primary and secondary umapi with a single directory group mapped to both in the normal way:

    - directory_group: "Directory Group"
      adobe_groups:
        - "Group Zero"
        - "secondary::Group Zero"

CLI Args were: 
--user-filter x --process-groups -t --adobe-only-user-action write-file x.csv

On python 3 I'm getting:

Traceback (most recent call last):
  File "G:/Repositories/Adobe/user-sync.py/user_sync/app.py", line 173, in sync
    begin_work(config_loader)
  File "G:/Repositories/Adobe/user-sync.py/user_sync/app.py", line 351, in begin_work
    rule_processor.run(directory_groups, directory_connector, umapi_connectors)
  File "G:\Repositories\Adobe\user-sync.py\user_sync\rules.py", line 195, in run
    self.process_strays(umapi_connectors)
  File "G:\Repositories\Adobe\user-sync.py\user_sync\rules.py", line 585, in process_strays
    self.write_stray_key_map()
  File "G:\Repositories\Adobe\user-sync.py\user_sync\rules.py", line 1152, in write_stray_key_map
    new_list = sorted(self.stray_key_map.keys())
TypeError: '<' not supported between instances of 'str' and 'NoneType'
2019-11-05 00:32:36 8152 INFO main - ========== End Run (User Sync version: 2.5.0) (Total time: 0:00:02) 

I think it will only throw if there were actually strays on the secondary (I didn't test that case).  So if it's just your primary, perhaps the string name of the secondary never was in the map at all. 

That said - totally agree, there's probably no real need to be sorting anyway.  I think it would make more sense to output a different file for each umapi which are named the same as the original but with the prefix.  If I were going to be using the results to do actions on console users, that's the format I imagine would be the most helpful.

adorton-adobe:
I think I figured it out.

1. Secondary orgs will only be included in the "stray key map" when `process_groups` is `True` and there is at least one Adobe group targeted to a secondary org (why??)
2. Only users that are a part of both the primary org and a secondary org will be written to the file.  This seems wrong to me.  The file should contain all Adobe-only users regardless of whether or not they are in the "primary".

I also think I found a bug - users in the secondary org (but only those with an account in the primary) will be removed from groups.  This shouldn't happen.  The `write-file` Adobe-only action should **not** make any changes to Adobe-only users.

For now, let's remove the sort() command from that function.  There's no reason to have it.  Users can sort the CSV however they want after generating it.  @sakonduri please modify the function to remove the sort altogether (and the comments that explain the sorting).

The other issues I brought up will be addressed at a later time.

user_sync/rules.py
```diff
@@ -1092,9 +1092,6 @@ def write_stray_key_map(self):
                 if not secondary_count:
                     fieldnames.append('umapi')
                 secondary_count += 1
-        # None sorts before strings, so sorting the keys in the map
-        # puts the primary umapi first in the output, which is handy
-        for umapi_name in sorted(self.stray_key_map.keys()):
```
adorton-adobe: @vossen-adobe don't we still need `umapi_name` here?  I don't think we want to remove the loop, just remove the call to `sorted()`

vossen-adobe: I combined it with the loop above instead of iterating over the same list twice

adorton-adobe: I don't see a commit for that change in this PR

vossen-adobe: It's the same as removing 1095-1097 - once the sorted line is gone, 1097 is identical to the for loop above it, so I just removed the line and comments.  There's no other changes to commit

vossen-adobe:
```
        for umapi_name in self.stray_key_map:
            if umapi_name != PRIMARY_UMAPI_NAME and self.get_stray_keys(umapi_name):
                if not secondary_count:
                    fieldnames.append('umapi')
                secondary_count += 1
        **for umapi_name in sorted(self.stray_key_map.keys()): <- just dropped this line**
            for user_key in self.get_stray_keys(umapi_name):
                id_type, username, domain = self.parse_user_key(user_key)
                umapi = umapi_name if umapi_name else ""
                if secondary_count:
                    row_dict = {'type': id_type, 'username': username, 'domain': domain, 'umapi': umapi}
                else:
                    row_dict = {'type': id_type, 'username': username, 'domain': domain}
                rows.append(row_dict)
```

adorton-adobe: Thanks, I see it now.  However we need that first loop to execute separately so we have a total `secondary_count` before we execute a the loop to write the CSV.  Otherwise we'll have undefined behavior because that variable governs which columns are included in the output.

vossen-adobe: Yes - that makes sense. Let me fix that!

Issue #517, opened 2019-07-30
Feature request: "groups not found" should be treated differently than a group with no users


michaelmernin:
A new feature would stop the UST from removing users from an admin console user group if the group they were being mapped from was not found.

For example, if an AD admin changed a security group name, the UST would no longer remove all user group membership on the admin console.

adorton-adobe:
Closing for now since we need to revisit the solution as per #521

Issue #503, opened 2019-06-20
Unexpected LDAP failure reading all users: invalid attribute type memberOf


n8felton:
```
2019-06-20 12:49:15 2875 CRITICAL main - Unexpected LDAP failure reading all users: invalid attribute type memberOf
```

I believe I have narrowed this down to this line of code:
https://github.com/adobe-apiplatform/user-sync.py/blob/04acb0dad6a03d86ca06912f0ba2738aa73a8e6c/user_sync/connector/directory_ldap.py#L313

If I manually change this value to the value of `group_member_attribute_name`, everything appears to work.

This value should likely not be hard coded and instead honor the `two_steps_lookup: group_member_attribute_name` value, while defaulting to `memberOf` if `group_member_attribute_name` is not configured.

AnxiousArchangel:
Similar results here also. In order to read group membership from an eDirectory (openLDAP) source, I had to add an alias of "memberOf" to the attribute "member" on groups. I then also defined two_steps_lookup: as "memberOf" to get it read. I expected that two_steps_lookup would not be necessary if "memberOf" is the default.

adorton-adobe:
This didn't come up in testing.  I will investigate further and try to reproduce it.

AcidRobot:
This is also happening to me.
Adding the alias fix from [AnxiousArchangel](https://github.com/adobe-apiplatform/user-sync.py/issues/503#issuecomment-505610743) fixed this issue for us.

adorton-adobe:
I can't reproduce it, most likely because I'm using a stock openLDAP container that doesn't appear to throw an error if I try to query memberOf (even though I don't have a schema installed to specify that field).

I agree this needs to be fixed, and I think the solution will be to decouple the member attribute name setting and the flag to enable two-step lookup.

bpoliakoff:
Seeing the behavior described above.  Here's a sanitized log (server = OpenLDAP) of the query that user-sync is sending:

Jul 25 16:50:03 servername local4.debug<167> slapd-slave[28065]: conn=6060086 op=4 SRCH base="cn=foo,ou=Group,dc=example,dc=com" scope=2 deref=3 filter="(objectClass=*)"
Jul 25 16:50:03 servername local4.debug<167> slapd-slave[28065]: conn=6060086 op=4 SRCH attr=member
[...]

PR #521, opened 2019-08-12
a solution for #517

Luci2015:
changes to both okta and ldap connector to critical stop processing when directory_group value is not found


Luci2015:
I've installed an openLDAP server and run the two-step-lookup with my version of the fix.
Results are good, as all expected members are being picked up and synced over to Admin Console.

Here is the content of my connector-ldap.yml:
```
username: "cn=Manager,dc=example,dc=com"
password: "the_pass"
host: "ldap://192.168.2.103"
base_dn: "DC=example,dc=com"
require_tls_cert: False
all_users_filter: (&(objectClass=person)(objectClass=top))
group_filter_format: "(&(|(objectClass=groupOfNames)(objectClass=posixGroup))(cn={group}))"
two_steps_lookup:
  group_member_attribute_name: "member"
  nested_group: True
user_email_format: "{mail}"
```
The user-sync-config.yml has this group mapping enabled, where `wwds` is in LDAP a group with only groupOfNames and top classes enabled:
```
  groups:
    - directory_group: "wwds"
      adobe_groups:
        - "active"
```
Structure of LDAP:




Here is the ldap initialisation picked from the log:
```
2019-09-05 15:44:39 80127 DEBUG ldap - ldap initialized with options: 
{
'group_filter_format': '(&(|(objectClass=groupOfNames)(objectClass=posixGroup))(cn={group}))', 
'all_users_filter': '(&(objectClass=person)(objectClass=top))', 
'group_member_filter_format': None,
'require_tls_cert': False,
'two_steps_lookup': {
      'group_member_attribute_name': 'member', 
      'nested_group': True 
  },
'string_encoding': 'utf8', 
'user_identity_type_format': None,
'user_email_format': '{mail}', 
'user_username_format': None, 
'user_domain_format': None, 
'user_given_name_format': '{givenName}',
'user_surname_format': '{sn}', 
'user_country_code_format': '{c}', 
'user_identity_type': 'federatedID', 
'search_page_size': 200, 
'logger_name': 'ldap',
'authentication_method': 'simple', 
'username': 'cn=Manager,dc=example,dc=com', 
'host': 'ldap://192.168.2.103', 
'base_dn': 'DC=example,dc=com', 
'two_steps_enabled': True
}
```
users are being found and added actions to add to **active** group on Adobe side:
```
2019-09-05 15:44:39 80127 DEBUG processor - Total directory users after filtering: 2
2019-09-05 15:44:39 80127 DEBUG processor - Group work list: {None: {'federatedID,usr0@example.com,': {'active'}, 'federatedID,usr1@example.com,': {'active'}}}
```

user_sync/connector/directory_ldap.py
```diff
@@ -310,13 +304,16 @@ def iter_users(self, base_dn, users_filter, extended_attributes):
         user_attribute_names.extend(self.user_email_formatter.get_attribute_names())
         user_attribute_names.extend(self.user_username_formatter.get_attribute_names())
         user_attribute_names.extend(self.user_domain_formatter.get_attribute_names())
-        user_attribute_names.append(six.text_type('memberOf'))
+
+        if not self.options['two_steps_enabled']:
+            user_attribute_names.append(six.text_type('memberOf'))
```
bajwa-adobe: `memberOf` shouldn't be requested from LDAP in any case. It is extra attribute which is ignored in most of LDAP directory but fails with above error on NetIQ eDirectory. This should be simply commented out from request.

vossen-adobe: I agree - but I think we shouldn't hardcode this anywhere - even in the conditional.  This line will break some openldap implementations.  If a user needs memberof, they can add it to the extension attributes?

adorton-adobe: We need to create an extra config where the attribute name can be specified. I don't think it would work to expect it to be in the extension config because this field has special meaning to the dynamic group mapping functionality.

Luci2015: The way I thought when I proposed the change was to separate AD from other LDAP implementations:
- AD will get memberOf because it's a default attribute of an user object
- the rest of the LDAP will go 'two steps lookup' with a special configuration since 'member' is a group object attribute that usually openLDAP has. The default ldap queries usually need a change when using openLDAP anyway.

At the time when I proposed the change I was not fully aware how the LDAP query was being formed and the change was to not risk breaking anything, but now I guess  I could just remove complectly the conditional and have the memberOf be part of the usual ldap filters in connector-ldap file.

Both ways, this should work.

adorton-adobe: `memberOf` is a separate issue and doesn't need to be addressed in this PR. We need to fix it in a separate PR that tracks with #503. Let's get this merged, then we can make a fix for #503.

Luci2015:
Since this PR dates back from Aug 2019, I wanted to know if the rest of the changes are needed or should I just close the entire PR. Thanks!

adorton-adobe:
@Luci2015 This is probably worth another look. If you can rebase against v2, resolve the conflicts in the LDAP connector and force-push your branch, I will take a look.

Luci2015:
This has been created more than 1 year ago. At that time the changes made sense, but since then the project morphed too much.
I do not have enough time to investigate what changes I need to apply, so I'll just close this PR.

PR #522, opened 2019-08-12
Updating example users-file.csv

n8felton:
Updating the rows to match the column header. `username` and `domain` were missing from the rows.

### DIFF
--- a/examples/csv inputs - user and remove lists/users-file.csv
+++ b/examples/csv inputs - user and remove lists/users-file.csv
@@ -1,4 +1,4 @@
 firstname,lastname,email,country,groups,type,username,domain
-John,Smith,jsmith@example.com,US,"AdobeCC-All",enterpriseID
-Jane,Doe,jdoe@example.com,US,"AdobeCC-All",federatedID
-Richard,Roe,rroe@example.com,US,"AdobeCC-All"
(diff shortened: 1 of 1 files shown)


Luci2015:
Nate,
The `username` and `domain` were intentionally left out because of a specificity of the API when creating the user accounts in an `email` vs `username` type of user login setting.
The createFederatedID will differ based on this flag.

For email setup JSON body starts with:
```
[{
  "user" : "jdoe@example.com",
  "do: [{.....}]"
}]
```
For username setup:
```
[{
  "user" : "jdoe",
  "domain" : "example.com",
  "do: [{.....}]"
}]
```
Inside Admin Console -> Settings -> Identity -> click inside a Directory in the list and check the `User login setting` value: email or username?

Most of the deployments I've seen have that value is set to `email` and the proposed change will fail the create action for them with a message such as:
`code: "error.user.email.invalid" message: "Email is invalid: jdoe"`
But the change may be good for `username` flag use.

Maybe the documentation should be a little more explicit on this part..

adorton-adobe:
I think it's better to leave them blank since email-type users are more common and we want the example to show the most common use case.  I'll make a note to update the docs in the right place.

PR #527, opened 2019-08-16
use jeepney 0.4 for py 3

vossen-adobe:
The ubuntu 3.6 build is failing across the board because the latest jeepney version (released Aug 11, 2019) seems to be broken for ubuntu.  Pex  reports:

wheel.install.BadWheelFile: No expected hash for file 'jeepney-0.4.1.dist-info/tmp/tmp4trcl86h/jeepney-0.4.1/LICENSE'

This fix overrides jeepney to previous version which works on all platforms.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,12 +56,14 @@ setup(name='user-sync',
           'umapi-client>=2.12',
           'click',
           'click-default-group',
-          'jeepney==0.4;python_version>="3"'
       ],
       extras_require={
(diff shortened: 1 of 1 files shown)


vossen-adobe:
@adorton-adobe  

That was extremely much more difficult than it should have been because windows and linux read the setup.py differently, but I finally found a solution that works on both, awkward as it is!

PR #539, opened 2019-10-01
Correct spelling on extensions_config.yml

sarlalian:
The file name for extensions_config.yml was spelled incorrectly.  This corrects that.

### DIFF
--- a/docs/en/user-manual/advanced_configuration.md
+++ b/docs/en/user-manual/advanced_configuration.md
@@ -202,7 +202,7 @@ configuration file in the `directory_users` section:
 
 ```
 directory_users:
-  extension: extenstions_config.yml  # reference to file with custom mapping information
+  extension: extensions_config.yml  # reference to file with custom mapping information
 ```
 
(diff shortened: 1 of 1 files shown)


sarlalian:
Updated the change.  Should be good to go now.

docs/en/user-manual/advanced_configuration.md
```diff
@@ -202,7 +202,7 @@ configuration file in the `directory_users` section:
 
 ```
 directory_users:
-  extension: extenstions_config.yml  # reference to file with custom mapping information
+  extension: extension_config.yml  # reference to file with custom mapping information
```
adorton-adobe: Sorry to nitpick, but to maintain the consistency I mentioned, we should specify the file as `extension-config.yml` (note the dash `-` instead of the underscore)

adorton-adobe:
I went ahead and made the change directly to v2 since it is a small documentation change.  It will be published on the next stable release.

Thank you for bringing this to my attention.

Issue #547, opened 2019-10-24
Avoid FederatedID/EnterpriseID creation if AdobeID exist


bhunut-adobe:
Currently UST will attempt to create FederatedID/EnterpriseID if AdobeID exists with the same email address but fail without error. UMAPI does not return an error.  This cause UST to attempt creation at every run.

To avoid confusion, UST should skip FederatedID/EnterpriseID creation if AdobeID exists with the same email.

Related to issue #511

PR #548, opened 2019-10-24
Added a rule to skip user creation if AdobeID exists with the same email

bhunut-adobe:
Added a rule to skip user creation if AdobeID exist with the same email as FederatedID/EnterpriseID
Example of the log output
```
2019-10-24 15:31:05 56505 WARNING processor - Skipping user creation for: federatedID,kbhunut@adobeccetest.com, - AdobeID already exists with kbhunut@adobeccetest.com
```

#547


user_sync/rules.py
```diff
@@ -954,6 +968,17 @@ def is_umapi_user_excluded(self, in_primary_org, user_key, current_groups):
...
+    def filter_adobeID_user(self, umapi_user):
+        id_type = self.get_identity_type_from_umapi_user(umapi_user)
+        if id_type == user_sync.identity_type.ADOBEID_IDENTITY_TYPE:
+            self.adobeid_user_by_email[normalize_string(umapi_user['email'])] = umapi_user
+
+    def is_adobeID_email_exist(self, email):
+        result = self.adobeid_user_by_email.get(normalize_string(email))
+        if result:
+            return True
+        return False
```
adorton-adobe: You can just cast the result of `self.adobeid_user_by_email.get()` to a bool and return that (which is essentially what you are doing)

e.g. 

```python
    def is_adobeID_email_exist(self, email):
        return bool(self.adobeid_user_by_email.get(normalize_string(email)))
```

bhunut-adobe: Ok. I will fix this. Admin Console would allow the federateID to be created if AdobeID with the same email exist if username does not equal to AdobeID email. I believe the admin console check is on the username field. Should I do validation on username instead of email?

adorton-adobe: Yes, if that is the case, I'd recommend it.  Have you tested that the UMAPI will allow users to be created in that manner?

bhunut-adobe: i think we should block the creation of FederateID if AdobeID exist with the same email. UMAPI allow this loop hole if username is different but Admin Console doesn't.

adorton-adobe: Sounds good.  Let me know when you've made the change and I'll test it.

PR #552, opened 2019-11-19
Post-sync and Sign Sync Connector

### LINKED ISSUE #551: Sign Sync Connector
We need to extend UST sync functionality to perform sync tasks for Adobe Sign. For customers with Adobe Sign Enterprise plans, users assigned to a Sign entitlement will automatically be granted access to Adobe Sign. These users will be assigned to the Default Group and granted Normal User privileges.

Some customers need the ability to assign users to non-default groups and may also need to assign Group or Account admin privileges to users. 

The Sign connector can use the Sign API to make these user account updates.

Since there may also be other Adobe products that may require similar product-specific sync workflows in the future, we will create a new type of connector - the **Post-Sync Connector**. This connector will operate on users that have been synced to the Admin Console after the main sync workflow of the UST has finished.

It will operate similarly to the UST itself - the UST's job is to maintain a consistent between an identity source and the Admin Console. The post-sync connector will maintain a consistent state between the Admin Console and a given Adobe product or service.

adorton-adobe:
* Implement the infrastructure for the post-sync connector
* Marshal post-sync data - a model representing the current state of users and group assignments in the Admin Console
* Implement the Sign sync connector - manage group assignments and admin privileges in Adobe Sign

Notes:

* Documentation is incomplete.  The manual page is done, but I still need to finish the tutorial page
* The new images in the user_manual directory are part of the tutorial.  I'll move them over (and probably delete those we don't need)

Fixes #551


docs/en/user-manual/post_sync_connector.md
```diff
@@ -0,0 +1,213 @@
...
+    adobe_groups:
+      - Adobe Sign Account Admins
+  - sign_role: GROUP_ADMIN
+    adobe_groups:
+      - Adobe Sign Group Admins
+```
+
+### Config Spec
+
+| key | type | required? | notes |
```
vossen-adobe: A clearer explanation of what the groups keys mean and how they are related would be helpful.  In addition, some discussion of what it means to use multiple sign orgs or umapis and how the sync works in that case.

user_sync/post_sync/connectors/sign_sync/client.py
```diff
@@ -0,0 +1,131 @@
...
+    def __init__(self, config):
+        for k in ['host', 'key', 'admin_email']:
+            assert k in config, "Key '{}' must be specified for all Sign orgs".format(k)
+        self.host = config['host']
+        self.key = config['key']
+        self.admin_email = config['admin_email']
+        self.console_org = config['console_org'] if 'console_org' in config else None
+        self.api_url = self.base_uri()
+        self.groups = self.get_groups()
+        self.default_group_id, = [_id for name, _id in self.groups.items() if name == self.DEFAULT_GROUP_NAME]
```
vossen-adobe: Unused field

user_sync/post_sync/connectors/sign_sync/client.py
```diff
@@ -0,0 +1,131 @@
...
+        This function validates that the SIGN integration key is valid.
+        :return: dict()
+        """
+
+        endpoint = self._endpoint_template.format(self.version)
+        url = 'https://' + self.host + '/' + endpoint
+
+        if self.version == "v6":
+            return requests.get(url + "baseUris", headers=self.header()).json()['apiAccessPoint'] + endpoint
+        return requests.get(url + "base_uris", headers=self.header()).json()['api_access_point'] + endpoint
```
vossen-adobe: We discussed on the call, but error handling in general will greatly improve the usability of the connector.  In this instance, the call to access a dictionary key 'api_access_point' fails with a cryptic KeyError if the api key is incorrect.

user_sync/post_sync/connectors/sign_sync/client.py
vossen-adobe: Unused field self.default_group_id

docs/en/user-manual/post_sync_connector.md
vossen-adobe: A clearer explanation of what the groups keys mean and how they are related would be helpful. In addition, some discussion of what it means to use multiple sign orgs or umapis and how the sync works in that case.

user_sync/post_sync/connectors/sign_sync/client.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def __init__(self, config):
+        for k in ['host', 'key', 'admin_email']:
+            assert k in config, "Key '{}' must be specified for all Sign orgs".format(k)
+        self.host = config['host']
+        self.key = config['key']
+        self.admin_email = config['admin_email']
+        self.console_org = config['console_org'] if 'console_org' in config else None
+        self.api_url = self.base_uri()
+        self.groups = self.get_groups()
```
vossen-adobe: The question of whether to call the sign API to query groups prior to running the sync tool.  Does it make more sense to call this method in the run() call so as to avoid 

(1) not breaking the normal sync if the call fails for some reason
(2) a potentially very long delay between when groups are fetched from sign and when they are actually used

I think its good to keep the base_uri() call since it provides immediate feedback if the api key is wrong.

OR - in contrast to the above, does it make sense to add some kind of post sync validation code to check the configuration of a post sync so we know the authentication will be good when the time comes.  Again, I'm just thinking in terms of large scale / long sync processes.  Might be overkill.

user_sync/post_sync/manager.py
```diff
@@ -0,0 +1,104 @@
...
+        return attributes
+
+    def run(self, post_sync_data):
+        """
+        run each entry from the module dict from __init__
+        :return:
+        """
+        for connector in self.connectors:
+            self.logger.info("Running module " + connector.name)
+            connector.run(post_sync_data)
```
vossen-adobe: Good opportunity for error handling to avoid hard stop if one of the post sync connectors errors out

adorton-adobe:
@vossen-adobe I've made the changes you've requested and have resolved the merge conflicts. Please take another look when you get a chance.

user_sync/config.py
```diff
@@ -386,6 +387,28 @@ def get_directory_extension_options(self):
...
+
+        try:
+            post_sync_modules = {m: self.get_dict_from_sources([connectors[m]]) for m in module_list}
+        except KeyError as e:
+            raise AssertionException("Error! Post-sync module " + str(e) + " specified without a configuration file...")
+
+        return {
+            'modules': post_sync_modules,
+        }
+
```
vossen-adobe, line 411: We need extra validation added here to prevent confusing errors.  See https://github.com/adobe-dmeservices/user-sync.py/pull/93

adorton-adobe: Cool, go ahead and merge it and I'll do some testing.

adorton-adobe: Looks good. I made one small tweak so we don't import a private member from the post sync connectors module.

examples/config files - basic/user-sync-config.yml
```diff
@@ -211,6 +211,19 @@ directory_users:
   # group_sync_options:
   #   auto_create: False
 
+# Post-sync connectors are enabled here
+# `modules` specifies by name the modules to be enabled
+# `connectors` specifies the location of each connector's config file
+#   NOTE: the key of each config file path must be a valid connector name,
+#         and must be represented in the `modules` list
+#  TODO create doc page and link it here
```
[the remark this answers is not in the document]
adorton-adobe: I added the link but it the page will not be live until we merge to master.

adorton-adobe:
@vossen-adobe Ready for another look

PR #556, opened 2020-01-10
jeepney fix no longer needed

vossen-adobe:
Removing the exception in setup.py, as builds now all pass without this, and the restriction is now causing problems with other libraries (secretstorage).

https://travis-ci.org/vossen-adobe/user-sync-fork/builds/635431553
https://ci.appveyor.com/project/vossen-adobe/user-sync-fork/builds/30033522

Example of failure:
https://travis-ci.org/vossen-adobe/user-sync-fork/jobs/635423817

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -58,9 +58,6 @@ setup(name='user-sync',
           'click-default-group',
       ],
       extras_require={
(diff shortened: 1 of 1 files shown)


vossen-adobe:
addressed by https://github.com/adobe-apiplatform/user-sync.py/pull/560

PR #578, opened 2020-03-03
Fix for 'invalid attribute type memberOf' at user-level LDAP request

bajwa-adobe:
Some OpenLDAP directories do not expect `memberOf` attribute in user-level LDAP request. It now reads memberOf attribute from `user_memberof_format` configuration. It is optional config and accepts `None`. The default value is `{memberOf}`.
#503

### DIFF
--- a/user_sync/connector/directory_ldap.py
+++ b/user_sync/connector/directory_ldap.py
@@ -77,6 +77,7 @@ class LDAPDirectoryConnector(object):
         self.user_given_name_formatter = LDAPValueFormatter(options['user_given_name_format'])
         self.user_surname_formatter = LDAPValueFormatter(options['user_surname_format'])
         self.user_country_code_formatter = LDAPValueFormatter(options['user_country_code_format'])
+        self.user_memberof_format_formatter = LDAPValueFormatter(options['user_memberof_format'])
 
         auth_method = options['authentication_method'].lower()
 
@@ -137,6 +138,7 @@ class LDAPDirectoryConnector(object):
         builder.set_string_value('user_given_name_format', six.text_type('{givenName}'))
         builder.set_string_value('user_surname_format', six.text_type('{sn}'))
         builder.set_string_value('user_country_code_format', six.text_type('{c}'))
+        builder.set_string_value('user_memberof_format', six.text_type('{memberOf}'))
         builder.set_string_value('user_identity_type', None)
         builder.set_int_value('search_page_size', 200)
         builder.set_string_value('logger_name', LDAPDirectoryConnector.name)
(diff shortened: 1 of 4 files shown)


user_sync/connector/directory_ldap.py
```diff
@@ -77,6 +77,7 @@ def __init__(self, caller_options):
         self.user_given_name_formatter = LDAPValueFormatter(options['user_given_name_format'])
         self.user_surname_formatter = LDAPValueFormatter(options['user_surname_format'])
         self.user_country_code_formatter = LDAPValueFormatter(options['user_country_code_format'])
+        self.user_memberof_format_formatter = LDAPValueFormatter(options['user_memberof_format'])
```
adorton-adobe: Let's change the name of the config option so we don't group it with the other `user_` config options. Those options govern how user attributes are mapped. Those attributes translate directly to Admin Console user attributes. The group membership attribute is a special case. It also shouldn't be a "format" variable since it doesn't make sense to make it formattable. So we can remove this line to create a formatter.

I suggest naming it something like `group_member_attribute`.

user_sync/connector/directory_ldap.py
```diff
@@ -137,6 +138,7 @@ def get_options(caller_config):
         builder.set_string_value('user_given_name_format', six.text_type('{givenName}'))
         builder.set_string_value('user_surname_format', six.text_type('{sn}'))
         builder.set_string_value('user_country_code_format', six.text_type('{c}'))
+        builder.set_string_value('user_memberof_format', six.text_type('{memberOf}'))
```
adorton-adobe: Since we aren't making this option formattable, we should remove the curly braces from the default value. Let's also change the default value to `'member'` since that is the default for the two-step attribute (which we will need to reconcile with this new feature).

user_sync/connector/directory_ldap.py
```diff
@@ -310,7 +312,7 @@ def iter_users(self, base_dn, users_filter, extended_attributes):
         user_attribute_names.extend(self.user_email_formatter.get_attribute_names())
         user_attribute_names.extend(self.user_username_formatter.get_attribute_names())
         user_attribute_names.extend(self.user_domain_formatter.get_attribute_names())
-        user_attribute_names.append(six.text_type('memberOf'))
+        user_attribute_names.extend(self.user_memberof_format_formatter.get_attribute_names())
```
adorton-adobe: We just need to append the value of our new config option here since we aren't doing formatting.

user_sync/connector/directory_ldap.py
adorton-adobe: On second thought, the default value should be `None` since this must be optional for basic use cases (e.g. when Dynamic group mapping and two-step lookup are not in use).

adorton-adobe:
@bajwa-adobe is this ready for another review?

bajwa-adobe:
Yes @adorton-adobe the PR is ready.

PR #583, opened 2020-03-13
Feature / Cert generation and key encryption

vossenv:
Documentation is here:
https://github.com/adobe-dmeservices/user-sync.py/blob/feature/native-certgen-enc/docs/en/user-manual/additional_tools.md

Allows to generate an x509 cert / keypair for umapi
Allows to encrypt/decrypt the private key using RSA algorithm
Added click functions to do this
Backwards compatible key decryption
Config file is not automatically updated to use encrypted key at this time - user will need to put in password. This can be handled more easily via the credential manager feature which is in process.


adorton-adobe:
@vossen-adobe is this ready for another review?

vossen-adobe:
@adorton-adobe Yep!

PR #584, opened 2020-03-13
Added whitelist for functions available inside of eval()

### LINKED ISSUE #588: Okta connector needs whitelist in eval()
Currently, the Okta connector has an eval() that takes a string value with no limitations of functions. This method should take a whitelist of builtin methods and a local dictionary (user) to prevent damaging code to be injected.

nickila:
Added whitelist of builtin Python functions to be available to the eval() function in the okta connector file.

fixes #588


nickila:
https://github.com/adobe-apiplatform/user-sync.py/issues/588

PR #586, opened 2020-03-13
removed python 2.7

### LINKED ISSUE #592: Remove Python 2.7
In preparation for the standalone EXE version of the tool, we no longer need to create builds that target Python 2.7.

sdhau:
removed python 2.7

fixes #592

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,8 +32,6 @@ setup(name='user-sync',
       description='Application for synchronizing customer directories with the Adobe Enterprise Admin Console',
       classifiers=[
           'Development Status :: 5 - Production/Stable',
-          'Programming Language :: Python :: 2.7',
-          'Programming Language :: Python :: 3.4',
           'Programming Language :: Python :: 3.5',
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -32,8 +32,6 @@
       description='Application for synchronizing customer directories with the Adobe Enterprise Admin Console',
       classifiers=[
           'Development Status :: 5 - Production/Stable',
-          'Programming Language :: Python :: 2.7',
-          'Programming Language :: Python :: 3.4',
           'Programming Language :: Python :: 3.5',
```
adorton-adobe: While we're at it can we remove the Python 3.5 tag? Technically it doesn't do anything because we don't publish this project to PyPI, but it is nice to keep it updated for the benefit of developers to hint at which Python versions we support.  Our CI builds all target 3.6 only in the 3.X series.

adorton-adobe:
@vossen-adobe seeing as how this update overlaps with the work we're doing in `build/exe` I'm going to go ahead an merge it,

PR #591, opened 2020-03-26
Document two-step lookup

### LINKED ISSUE #590: Two-step lookup needs documentation in manual
It doesn't look like we've documented two-step lookup functionality in the UST manual. We should add a section for it in **Advanced Configuration**.

adorton-adobe:
Add documentation for two-step lookup to the **Advanced Configuration** section of the UST manual.

fixes #590

### DIFF
--- a/docs/en/user-manual/advanced_configuration.md
+++ b/docs/en/user-manual/advanced_configuration.md
@@ -860,6 +860,29 @@ created on those organizations if `auto_create` is enabled.
 Refer to [Accessing Users in Other Organizations](https://adobe-apiplatform.github.io/user-sync.py/en/user-manual/advanced_configuration.html#accessing-users-in-other-organizations)
 for more information.
 
+## Two-Step Lookup
+
+Some LDAP systems may not support a `memberOf` overlay the way other systems like Active Directory do. This can make it
+impossible for the LDAP connector to use its group-based queries to get members for particular groups in an LDAP system.
+
+The `two_steps_lookup` option in `connector-ldap.yml` can be used to overcome this limitation. With two-step lookup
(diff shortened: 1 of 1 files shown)


adorton-adobe:
@bhunut-adobe no idea why the build is failing. I'll look into it. We may need to wait until we merge the standalone build since it will get rid of 2.7 (the build is failing on 2.7 for some reason). Please review when you get a chance since I want to make sure my information around two-step lookup is accurate. Once you approve I'll merge it when the build is sorted out.

docs/en/user-manual/advanced_configuration.md
```diff
@@ -860,6 +860,29 @@ created on those organizations if `auto_create` is enabled.
...
 for more information.
 
+## Two-Step Lookup
+
+Some LDAP systems may not support a `memberOf` overlay the way other systems like Active Directory do. This can make it
+impossible for the LDAP connector to use its group-based queries to get members for particular groups in an LDAP system.
+
+The `two_steps_lookup` option in `connector-ldap.yml` can be used to overcome this limitation. With two-step lookup
+enabled and configured, the sync tool will query all users (with respect to `all_users_filter`) and look for a specific
+user attribute to get group membership information. Users queried in this manner will be synced according to the `--users`
```
bhunut-adobe, line 870: @adorton-adobe
Not sure if the wording here is correct. Since we don't pull all the users first then filter.
Two steps lookup process:
1) Query Specified Group and retrieve group membership from specified "group_member_attribute_name" value. This will return only DN Value
2) Verify if the users are within the base dn
3) Lookup user dn by using all_users_filter filter to return Person Object.
4) Move on to the rest of connector ldap stuff.

The rest of the doc look good.

PR #593, opened 2020-04-01
Standalone EXE builds

### LINKED ISSUE #479: PyInstaller Build
For security reasons, some customers are unable to install a full Python environment on any of their servers, making it impossible to run the User Sync Tool.  The concern is that interpreter presents a vector for arbitrary script execution.

PyInstaller can produce a standalone executable for any of our supported platforms that will embed a Python interpreter into the executable itself.

If a PyInstaller build proves to be feasible and reproducible, it may be good to think about using it to make our builds instead of pex in the future (post-3.0 since it would radically break backwards compatibility).  For now, any builds we make will be done on a parallel fork.  If we merge this feature pre-3.0, it will be an optional build target.  pex will continue to be the main and default build target.

The major challenge to building with PyInstaller is an upstream dependency of `keyring` which isn't compatible with PyInstaller's dependency analysis process.  @vossen-adobe has been investigating this issue and has a potential solution.

This build will also need some security-related build options including:

* Option to disable extension config
* Option to disable shell execution in config file links

This will require a framework to manage feature flags.  The feature flag manager will require a resource manager because a flag config data file will need to be packaged in the build to control these features.

adorton-adobe:
Moving from pex-based builds to standalone executables. The sync tool can now run with no external dependencies. A Python interpreter is not required. This means that we no longer have to support Python 2.7. Users that want to upgrade to 2.6.0 once it is released will just need to deploy the executable version and tweak any invocation scripts they may be using.

Summary of changes:

* Update Makefile, setup.py and CI config to only make standalone builds
* Remove Python 2.7 builds
* Create a build variant that disables extension support
* Update docs to remove all pex references and base examples on executable
* Remove irrelevant Python references from docs since a Python interpreter is no longer required
* Update build instructions in README.md for Windows and Linux

fixes #479


adorton-adobe:
@vossen-adobe Do you mind taking a look at this? I think the code and build changes have been pretty well-vetted at this point, but take a quick look and let me know if anything stands out. And let me know if the documentation changes look good.

PR #603, opened 2020-05-05
skip user from the user sync tool process

sriramsrini125:
Changes has been done in rules.py to remove the user from user sync tool process based on the condition in extension-config.yml file. skip_user() need to be called if the condition satisfied to remove the user in extension-config.yml


user_sync/rules.py
```diff
@@ -361,14 +361,15 @@ def read_desired_user_groups(self, mappings, directory_connector):
         extended_attributes = options.get('extended_attributes')
 
         directory_user_by_user_key = self.directory_user_by_user_key
+        filtered_directory_user_by_user_key = self.filtered_directory_user_by_user_key
```
vossenv, line 364: Where did this come from?

vossen-adobe:
Closing after updating and re-making: https://github.com/adobe-apiplatform/user-sync.py/pull/647

PR #608, opened 2020-05-26
Ssl cert verify update

vossen-adobe:
Update cert verification:

- Move to cover all https connections, not just UMAPI since it doesn't make sense to apply this setting on a per-connector basis.  Now lives as CLI arg and invocation defaults that apply to all connections.
- Doesn't work with Okta - since okta's requests are external
- Disables verbose warning for every request, and instead displays warning at start of sync
- Raises error with info to switch to new method if found in connector-umapi


user_sync/app.py
```diff
@@ -353,6 +359,17 @@ def begin_work(config_loader):
     directory_groups = config_loader.get_directory_groups()
     rule_config = config_loader.get_rule_options()
 
+    ssl_cert_verify = rule_config.get('ssl_cert_verify', True)
```
adorton-adobe, line 362: While we normally prefer the `get()` dict method, I don't think it makes sense to use it here. Since the `ssl_cert_verify` key originates from the `invocation_defaults` dict, we are safe to assume the key is present here. If the key doesn't exist for some reason, we want to get the error instead of silently providing a default value.

user_sync/app.py
```diff
@@ -353,6 +359,17 @@ def begin_work(config_loader):
...
+    user_sync.connector.umapi.UmapiConnector.ssl_cert_verify = ssl_cert_verify
```
adorton-adobe, line 363: If we modifythe `ssl_cert_verify` class property here, it will change that value for all `UmapiConnector` objects, and we may not always want to do that in the future. I recommend passing `ssl_cert_verify` as a new parameter to the `UmapiConnector` constructor and setting it as an object attribute.

user_sync/app.py
```diff
@@ -353,6 +359,17 @@ def begin_work(config_loader):
...
+    user_sync.post_sync.connectors.sign_sync.SignClient.ssl_cert_verify = ssl_cert_verify
```
adorton-adobe, line 364: See my previous note - modifying a class property can have unexpected and often undesirable side effects.

user_sync/app.py
vossen-adobe: Sure - I can make this change.  I intentionally chose class properties because setting this value should be consistent across all the affected modules, regardless of how they  are instantiated.  It doesn't make sense to apply it to a single object but not another one.  Sync would not succeed unless they were all disabled (the driving force for this PR).  Using class properties was the simplest way to accomplish this.  I will rework it as an arg instead for primary/secondaries, sign sync, and console connector.

adorton-adobe: Thanks, It isn't as much about how we do things now vs. how things may change in the future. In general, changing things that can affect the scope beyond a single object or variable can create bugs that are difficult to track down.

PR #616, opened 2020-06-15
Rule process refactor

sriramsrini125:
Removed Rules.py and added the user_sync.py file and moved the code from rules.py to user_sync.py under the engine module


vossen-adobe:
Moved https://github.com/adobe-dmeservices/user-sync.py/pull/113

PR #617, opened 2020-06-15
Sign post-sync bug fix - email casing difference between umapi and sign users

ddybvig-perf:
Fixes issue #632 

The `update_sign_users` method in the SignConnector was skipping sign users if their email had different casing from the UMAPI users. Since the Sign API returns all lowercase emails, this was causing unwanted skips. For example, the loop would start with a sign user "user@example.com." Then it would look up that email from the UMAPI users dictionary. If the corresponding UMAPI user was "User@example.com," there would be no match and the should_sync method would be passed None for the Sign user, thus returning false, thus skipping the rest of the update.

The fix in SignConnector was just a `.lower()`. This PR also includes a unit test which checks the lookup of a Sign user against a UMAPI user with an email address that is identical except for a capital letter.

### DIFF
--- a/tests/test_post_sync.py
+++ b/tests/test_post_sync.py
@@ -68,10 +68,8 @@ def test_update_sign_users(mock_client, mock_get, example_user):
         r.status_code = 200
         r._content = json.dumps(data).encode()
         return r
-    #     user_list = mock_response({'userInfoList': [{"userId": "123"}, {"userId": "456"}]})
     user_list = mock_response({'userInfoList': [{"userId": "123"}]})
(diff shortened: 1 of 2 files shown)


adorton-adobe:
@ddybvig-perf  Please create an issue for this and link it to this PR by adding a "fixes #xxx" line to the description. We need tracking issues for all PRs so we can track and triage bugs appropriately.

ddybvig-perf:
@adorton-adobe Done. Let me know if I made any errors or need to do anything else.

adorton-adobe (approved):

PR #619, opened 2020-06-17
Product profile sign bug

### LINKED ISSUE #645: Product profile names don't match due to appended extra characters
**Description**
Product profiles don't always match when they should due to additional characters being appended to the end of a product profile name in user groups in the umapi response. 

**Steps to reproduce**
* Relevant config options
* Command-line options used
* Specific details around identity source, group mapping, etc.

**Expected behavior**
product profile names should match even if one has a certain string of characters appended to the end if it matches a given regex.

**Screenshots**

**Environment**
 - UST version:
 - OS type and version:
 - Any other relevant info:

nickila:
Product profiles don't always match when they should due to additional characters being appended to the end of a product profile name in user groups in the umapi response. This fix will remove the appended characters before they are compared. A unit test was created in test_sign_client.py for should_sync to verify the bug fix.

fixes #645

### DIFF
--- a/tests/test_sign_client.py
+++ b/tests/test_sign_client.py
@@ -1,0 +1,18 @@
+from user_sync.post_sync.connectors.sign_sync.__init__ import SignConnector
+
+
+def test_should_sync():
+    client_config = {
+        'console_org': None,
+        'host': 'api.na2.echosignstage.com',
+        'key': 'examplekey',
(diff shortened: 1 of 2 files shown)


adorton-adobe:
After some internal discussion, it looks like this change isn't necessary since `entitlement_groups` can specify more than one group (profile or user group) to define Sign entitlements for the connector.

PR #621, opened 2020-06-30
Test mode for Sign sync

### LINKED ISSUE #637: Test mode for Sign sync
Current Functionality: Right now the user sync has a test mode option; however, the post sync connector is disabled in this case because if the sync tool doesn't make any real changes then the actions for Sign sync would not work.

The above functionality makes sense and should not be changed; however, as a <user> I need <Sign sync to have it's own test mode> where it will state the changes it plans to make but not execute them. Therefore, we will not know the actual response from Sign when using test mode, but we will at least be able to see a record of the changes we plan to make.  UST will still need to be run in live mode for this to make sense.

mbonish:
fixes #637 
Sign sync to has its own test mode where it will state the changes it plans to make but not execute them. sign_test_mode displays post_sync output without performing any updates. Users and user_groups will not be updated in the Sign Dashboard, but users and user groups to be changed will be displayed.   

Updated connector-sign-sync.yml (comments for clarity and default value if sign_test_mode no) to the following:
 #Test_mode presents post_sync output without performing any updates of users or user groups to the Sign Dashboard
    # sign_test_mode possible values are yes or no. (default no)
    # sign_test_mode is independent of user_sync test mode.
    # Post Sync option does not run if user_sync tool is in test mode
    sign_test_mode: no

Sign_test_mode is independent of user_sync test mode. UST will still need to be run in live mode for this to make sense. The post sync connector is disabled in test mode.

### DIFF
--- a/examples/sign/connector-sign-sync.yml
+++ b/examples/sign/connector-sign-sync.yml
@@ -10,7 +10,11 @@ sign_orgs:
     key: "Sign API key"
     # Sign account associated with admin_email value will be skipped
     admin_email: "Admin user associated with key e.g. user@example.com"
-
+      # Test_mode presents post_sync output without performing any updates of users or user groups to the Sign Dashbaord
+      # sign_test_mode possible values are yes or no. (default no)
+      # sign_test_mode is independent of user_sync test mode.
+      # Post Sync option does not run if user_sync tool is in test mode
+      sign_test_mode: no
 # User groups to sync to Adobe Sign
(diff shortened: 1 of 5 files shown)


adorton-adobe:
@mbonish Please create an issue for this and link it to this PR by adding a "fixes #xxx" line to the description. We need tracking issues for all PRs so we can track and triage bugs appropriately.

adorton-adobe:
Addressing this in Sign Phase II

PR #623, opened 2020-07-24
Bug Fix-adobe-only-user-list with post sync

### LINKED ISSUE #638: Adobe only user list broken with post sync
If Adobe only user list is used with user sync 2.6 (current version) a nonetype exception is thrown because the umapi post sync data fails to handle this case properly

mbonish:
fixes #638 
If Adobe only user list is used a none type exception was thrown with post sync. Fix was to handle exception  if  adobe-only-users- list used. Fixed to handle exception: "Post-Sync Connectors only support "sync" strategy".


adorton-adobe:
@mbonish Please create an issue for this and link it to this PR by adding a "fixes #xxx" line to the description. We need tracking issues for all PRs so we can track and triage bugs appropriately.

mbonish:
Should be good now. Sorry for the delay.

vossen-adobe:
@adorton-adobe this is still relevant and could be merged or closed (only applies to post sync)

adorton-adobe (approved):

PR #630, opened 2020-09-09
V2 credential manager multi directory

sriramsrini125:
Features of credential manager and multi directory has been added to v2


vossen-adobe:
Where is this coming from? I don't think we are ready yet

vossen-adobe:
We haven't merged cred manager, and multi directory will need to follow that.

vossen-adobe:
Closing

PR #647, opened 2020-10-05
Allow extension config to skip a user during sync

### LINKED ISSUE #646: Allow sync to skip users from extension
**Is your feature request related to a problem? Please describe.**
There are many instances in which it is useful to "skip" a user during sync.  For example, the most common one would occurs when filtering users by directory attributes (e.g., group by department but don't sync users not in departments).  More detail is in #466, which we closed to add a better solution.

**Describe the solution you'd like**
We'd like to be able to skip a user during sync (i.e., completely remove them from the list and do not sync them against the console) from the extension.

**Describe alternatives you've considered**
Directory filtering is insufficient using LDAP queries.  Instead, we have frequently just changed a user's email to be invalid within the extension, causing the UMAPI to skip them.  This generates a lot of errors, and is not a good way to accomplish the task.

**Additional context**
See #466

vossen-adobe:
## Summary
* Remaking this PR from #603 to make it current and isolated.  Basically, we provide a locally scoped skip_user() function by reference to the extension config.  Calling this within the extension will allow it to reach back into the code and pop the user from the required places in order to remove them completely from remaining sync process.

## Testing Steps
* 

Fixes #646


vossen-adobe:
need a unit test for this

adorton-adobe:
Let's put this on hold for now.

PR #650, opened 2020-10-07
removeFromOrg not functionning correctly in some cases

### LINKED ISSUE #649: case sensitve username value makes removeFromOrg actions impossible
**Description**
case sensitve username value makes removeFromOrg actions impossible

**Steps to reproduce**
- create an account in Admin Console with email <> username, both email fomated values
- make the username to have at least an uppercase letter
- run the UST with '''--adobe-only-action-remove''' so that the new account shows as Adobe-only and the tool to create the remove action
- observe the "user" value in the json is the one for Username instead of Email. This action fails on UMAPI side. This remove action will appear on all the rest of the UST runs.
* Relevant config options
* Command-line options used
user-sync.exe --adobe-only-action-remove
* Specific details around identity source, group mapping, etc.

**Expected behavior**
the JSON should show the emaill value for "user" instead of username; the extra account gets to be removed at the end of the action

**Screenshots**

**Environment**
 - UST version: 2.5.0+
 - OS type and version: all
 - Any other relevant info:

Luci2015:
## Summary
* removeFromOrg actions will not work if there is a letter case difference between the Admin Console Username field and the internal dict key used to store information about associated email, in the case where both Email and Username fields contain different email formatted values. 
* The fix is to force lowercase values for usernames before these are getting added as keys to the dict 'self.email_override'

## Testing Steps
* In Admin Console create a 'supplementary' account that has as an Username an email with some uppercase letters.
* run the UST with '--adobe-only-user-action remove' policy
* you will observe that the JSON body for the removeFromOrg action contains current username value for the "user" statement; normally this is recorded as a failure on UMAPI side and on each run the remove action will be shown and never actually done
* UMAPI works by targeting the account by its email, not its username
* now make a build containing the provided fix and you will observe "user" picking up correctly the email value and the remove actually happening in Admin Console

Fixes #649

### DIFF
--- a/user_sync/post_sync/connectors/sign_sync/client.py
+++ b/user_sync/post_sync/connectors/sign_sync/client.py
@@ -98,7 +98,7 @@ class SignClient:
         """
         retry_nb = 1
         waiting_time = 20
-        while retry_nb < 6:
+        while retry_nb < 5:
             try:
                 waiting_time *= 3
                 self.logger.debug('Attempt {} to call: {}'.format(retry_nb, url))
@@ -112,7 +112,7 @@ class SignClient:
                     raise AssertionException('')
             except Exception as exp:
                 self.logger.warning('Failed: {}'.format(exp))
-                if retry_nb == 5:
+                if retry_nb == 4:
                     raise AssertionException('Quitting after 3 failed retry attempts')
                 self.logger.warning('Waiting for {} seconds'.format(waiting_time))
                 time.sleep(waiting_time)
--- a/user_sync/post_sync/connectors/sign_sync/client.py
+++ b/user_sync/post_sync/connectors/sign_sync/client.py
@@ -1,7 +1,10 @@
 import logging
-import requests
 import json
+import time
+
 from user_sync.error import AssertionException
+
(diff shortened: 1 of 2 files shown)


adorton-adobe:
Thank you for submitting this.

Will you please fill out the template? The testing steps are especially critical since they help the reviewer know how to verify that the change is doing what it should be doing.

Also - please update the PR title to concisely summarize the change. This PR title typically goes into release notes, so it's important that it makes sense to somebody reading those notes. The reference to the bug this fixes should go where it's stated in the template ("fixes #xxx"). That way it links the bug in a way that Github understands and if we merge the PR, the bug will be automatically closed.

adorton-adobe:
@Luci2015 - I see some changes that appear to be related to the Sign connector. Can you move those commits to a separate branch so this PR just covers the change described in the notes? Or alternatively, we can close this PR and you can open a new one that has this change isolated on a different branch.

Luci2015:
I'll close this PR entirely. Apparently the tests have to be reviewed as well - the 3rd attempt should be expected to fail with an  exception followed by a stop.

PR #662, opened 2020-12-02
Add additional guidelines to the scheduling page

lesavage:
Updating the scheduling documentation to include some additional guidelines to encourage clients to choose a sync time that does not overlap with the majority of clients.


adorton-adobe (approved):

adorton-adobe:
I've also merged this commit directly to master so the doc will get updated prior to the next release.

Issue #681, opened 2021-03-04
Make sign sync asynchronus


vossen-adobe:
**Is your feature request related to a problem? Please describe.**

The opportunity to significantly enhance the speed of sign sync exists.  We can multithread the requests using asyncio

PR #682, opened 2021-03-04
Sign Post Sync async and retry handling

vossen-adobe:
## Summary
* Add asyncio and retry to sign sync 
* Add options to set concurrency level (number of connections) and batch size 
* Add options for retry

The request set is broken down into batches, and each batch is executed via asyncio, using semaphore to control the max number of threads.  This ensures we don't over allocate memory by queueing all requests at once (and more than doubling the memory space for users), which is actually significant for high numbers of users.  

Tests run up to 40x concurrency on 150k users, for both user retrieval and updates.  Performance increases measured to be on the order of N^0.90, where N is the semaphore.

Also implement retry from https://github.com/Luci2015/user-sync.py/blob/v2/user_sync/post_sync/connectors/sign_sync/client.py#L125

*Note retry output looks confusing for async, since many threads can have their own retry counts.  Nevertheless, it does work correctly.

https://github.com/adobe-apiplatform/user-sync.py/issues/681


adorton-adobe:
I'm a bit confused about how the batch size option works and how it works with the concurrency level.

vossen-adobe:
@adorton-adobe to use asyncio, we need to create a list of tasks for it to handle.  Instead of creating all the tasks at once, the batch size can limit the total objects in that list.  All tasks in the batch are executed async using N threads until it is empty.

The reason is because creating a list of coroutines for users takes up a lot of memory if you have a lot of users - since on top of this we already have several lists of (in this case) 150k user objects (directory and umapi users).  I added this because my worst case test of 1 gig ram was running out of memory trying to run.

vossen-adobe:
@adorton-adobe do we still want this feature in Post Sync, or just close since included in phase 2?

examples/sign/connector-sign-sync.yml
```diff
@@ -11,6 +11,16 @@ sign_orgs:
...
+connection:
+  # Number of allowed concurrent requests (higher is faster, but consumes more bandwidth and memory)
+  request_concurrency: 5
+  # Number of requests to queue at one time.  Reduce if memory usage is too high.
+  batch_size: 10000
+  # Number of times to retry failed requests
+  retry_count: 5
+  # Timeout for requests in seconds
+  timeout: 120
+
```
adorton-adobe, line 23: I don't know if it makes sense to document these options. We should make them available, but I think the defaults will work for most customers. What do you think about making them undocumented? We may mention them in documentation in the future, but for now we can work with individual customers to tweak these settings if they experience problems.

adorton-adobe (changes requested):
Looks good, but I think we should consider removing the connection control options from the example config. I think they will be confusing for most users, and the defaults seem reasonable. If customers experience problems, we can recommend changes on an as-needed basis. We can also document these as advanced options in the manual if we want.

adorton-adobe (approved):

PR #755, opened 2021-12-17
Update sign.py

### LINKED ISSUE #753: Sign sync for users with different letter case for the email
**Description**
A clear and concise description of the issue.
Sign sync for users with different letter case between directory_user and sign_user. The mismatch makes the tool not find the directory user and skips it.

**Steps to reproduce**
- Create an account with uppercase email in Admin Console (GPS Org)
- add the account to a group mapped for Sign sync
- run sync for Sign
- the account gets skipped

**Expected behavior**
- the account should get processed

**Environment**
 - UST version: 2.7.0
 - OS type and version: all
 - Any other relevant info: Email@domain.com below is the Email field value in Admin Console for this account
2021-12-17 10:00:53 82270 INFO sign_sync - User Email@domain.com not present and will be skipped.
 
looks like the dict to match the sign extraction vs the db contains lower case letters for email key

Luci2015:
## Summary
* difference in letter case between Sign all users extraction and individual user information extraction induces an error of key mismatch between directory_user expected to be matched against extracted sign_user.

## Testing Steps
* create an account that has upper case letters in Admin Console (GPS Org) and entitle it to Sign
* Let entitlement finish so that the account appears in Default Group
* Assign the account in admin console to a mapped sign goup inside the Sign config file
* run as usual with sign-sync

I only tested for ACTIVE users but I noticed there is similar code for the INACTIVE ones, so I added the fix there as well.

Fixes #753

### DIFF
--- a/user_sync/engine/sign.py
+++ b/user_sync/engine/sign.py
@@ -154,13 +154,13 @@ class SignSyncEngine:
         dir_users_for_org = {}
         self.total_sign_user_count += len(sign_users)
         self.sign_users_by_org[org_name] = sign_users
-        for _, directory_user in directory_users.items():
+        for directory_user_key, directory_user in directory_users.items():
 
             if not self.should_sync(directory_user, org_name):
                 continue
 
-            sign_user = sign_users.get(directory_user['email'].lower())
-            dir_users_for_org[directory_user['email'].lower()] = directory_user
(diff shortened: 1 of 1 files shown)


Luci2015:
This fix seems to work only for the initial sync that creates the cache. Need to investigate sync from cache now

adorton-adobe (approved):

PR #774, opened 2022-03-21
Simplify identity type override for ESM trustees

### LINKED ISSUE #767: UST 2.7.0 does not handle BusinessIDs in ESM child consoles
**Description**
Known issue added here for visibility
UST does not handle correctly the exsiting BusinessIDs in child consoles when it picks up the accounts to sync from a different console

**Steps to reproduce**
* usual UST setup with adobe_console as source of federatedID accounts
* user-sync.exe
* the domain is trusted in ESM console from the source 'main' console
* existing businessIDs get remove action

**Expected behavior**
existing businessIDs that match the federatedID email account synced from source console should be managed based on group mapping

**Environment**
 - UST version: latest
 - OS type and version: all
 - Any other relevant info: the problem comes from the difference between the key that comes in from source console vs the key obtained when extracting from target console.

adorton-adobe:
## Summary
* When `uses_business_id` is enabled, just override the identity type without requiring a primary target
* Update docs

## Testing Steps
* Sync to a T2E trustee with no primary target

Fixes #767

### DIFF
--- a/docs/en/user-manual/advanced_configuration.md
+++ b/docs/en/user-manual/advanced_configuration.md
@@ -204,8 +204,10 @@ enterprise:
 uses_business_id: True
 ```
 
-This essentially overrides the Business ID user type to the type of the user from the primary target, ensuring that
-the full user lifecycle of the user on the secondary target is managed.
+This setting overrides the identity type of all users in the target to the identity type specified in the main sync config file.
+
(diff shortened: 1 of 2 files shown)


adorton-adobe:
@Luci2015 This should resolved #767 but needs more testing. Please let me know if you can take a look.

PR #781, opened 2022-05-03
Update CredentialManager to accommodate changes to ConfigLoader and ConfigFileLoader

### LINKED ISSUE #780: Credentials ConfigLoader updates
Currently some tests from test-credentials.py are failing because there have been some changes to the ConfigFileLoader and ConfigLoader classes, namely that the load_root_config method is no longer static and the as_list method needs to be called from the UMAPIConfigLoader child class. The calls to these methods in the CredentialManager class need to be updated so that the tests will pass and we maintain functionality.

ddybvig-perf:
## Summary
This pull request updates the calls to the ConfigFileLoader.load_root_config() and the ConfigLoader.as_list() methods. load_root_config() is no longer a static method, so this PR creates an instance of the ConfigFileLoader class within CredentialManager to then call that method. The ConfigLoader parent class no longer has a method as_list() since that method has been moved the UMAPIConfigLoader child class, so I made the appropriate changes for that as well.

There was one other test in test-credentials.py that was failing because it was passing connector-umapi.yml as a root config instead of the expected root config, user-sync-config.yml. I think this was just a mistake in how the test was written, so I changed it to use the correct file and it now passes.

## Testing Steps
I did not create any new tests, but I did change one of them as described above.

Fixes #780

### DIFF
--- a/tests/test_credentials.py
+++ b/tests/test_credentials.py
@@ -123,7 +123,7 @@ def test_credman_retrieve_revert_valid(test_resources, modify_config):
 
 def test_credman_retrieve_revert_invalid(test_resources, modify_config):
     umapi_config_file = modify_config('umapi', ['enterprise', 'priv_key_path'], test_resources['priv_key'])
-    credman = CredentialManager(umapi_config_file)
+    credman = CredentialManager(test_resources['umapi_root_config'])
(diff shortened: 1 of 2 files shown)


adorton-adobe:
Credential manager on hold for now.

PR #787, opened 2022-05-31
Update test_dcmanager.py

### LINKED ISSUE #786: dc_manager fixture update
**Description**
The tests in test_dcmanager.py currently fail in the v2-multi-dir-cred branch because of the changes that have been made to the config loader classes. The dc_manager fixture currently uses a ConfigLoader object to load in the main config file since that was the architecture within the tool when the test was written. Instead of a ConfigLoader object, we should use a UMAPIConfigLoader to reflect the current architecture.

ddybvig-perf:
## Summary
The dc_manager fixture needs to use a UMAPIConfigLoader object instead of a ConfigLoader to reflect updates that have been made to architecture elsewhere in the Sync Tool. There is also a test that uses mock.patch that needs to point to the path of the UMAPIConfigLoader instead of the ConfigLoader.

## Testing Steps
I updated the fixture and the mock.patch decorator to use UMAPIConfigLoader, then verified that all tests within test_dcmanager.py now pass.

Fixes #786

### DIFF
--- a/tests/test_dcmanager.py
+++ b/tests/test_dcmanager.py
@@ -4,7 +4,8 @@ from unittest import mock
 
 import pytest
 
-from user_sync.config.common import ConfigLoader
+from tests.conftest import default_args
+from user_sync.config.user_sync import UMAPIConfigLoader
 from user_sync.dcmanager import *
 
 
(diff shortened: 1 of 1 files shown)


adorton-adobe:
I merged this directly in [b9079a5](https://github.com/adobe-apiplatform/user-sync.py/commit/b9079a526d42e3df52341f4380bc1ebd066a13af)

PR #792, opened 2022-07-04
fixes #782 - avoid running the Sign extraction in between LDAP connection initialisation and LDAP extraction

### LINKED ISSUE #782: Sign-Sync 2.7.3 error - Unexpected LDAP failure reading group info: session terminated by server
**Description**
When syncing users from LDAP to Sign, this error is generated every day when the cache is refreshed:
 "CRITICAL main - Unexpected LDAP failure reading group info: session terminated by server"

**Steps to reproduce**
The error is replicable only when the following circumstances are met:
* the identity source is LDAP
* the number of Sign users is high (over 7000) and retrieving all Sign users lasts for ~2hours
* the Sign-Sync Tool is executed after more than 24 hours from the previous cache refresh
* (If the cache is fresh, the tool runs successfully) 

**Expected behavior**
The tool should run properly, like it does when the cache is fresh

**Environment**
 - UST version: 2.7.3

Luci2015:
## Summary  

Where Sign users' number is big (>3000), it takes a while to extract all of the users from Sign. In cases observed, the initial daily refresh of the cache takes 1-2 hours to finish with the Sign extraction.
UST has a logic to initialise the connector first, so LDAP connection is the first operation, then follows the Sign extraction, then follow the LDAP accounts' extraction, which in some conditions fails because there is a timeout between initial connection and the actual extraction.

## Testing Steps  
Have 7000 accounts in Sign, or simulate that the Sign extraction takes 1 hour
Run the Sign sync. After one hour the LDAP extraction should work without erroring.  

Fixes #782

### DIFF
--- a/tests/test_certgen.py
+++ b/tests/test_certgen.py
@@ -33,7 +33,7 @@ def test_create_key(key):
 def test_create_cert(random_subject, key):
     cert = create_cert(random_subject, key)
     cert_dict = {i.oid._name: i.value for i in cert.subject}
-    for k, v in six.iteritems(cert_dict):
+    for k, v in cert_dict.items():
         assert random_subject[k] == v
     random_subject['countryName'] = 'usa'
     with pytest.raises(AssertionException):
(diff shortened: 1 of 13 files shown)


adorton-adobe:
My latest update to `setup.py` created a merge conflict. Can you please update your branch and I'll get this merged?

PR #809, opened 2022-12-05
Update umapi.py

### LINKED ISSUE #808: users not processed correctly when username and email have different domains
**Description**
Users are not processed correctly when username and email have different domains when removing from file.

**Steps to reproduce**
When using write-to-file the tool exports [type,username,domain]
When removing from file (--adobe-only-user-list users-to-delete.csv --adobe-only-user-action delete) the username value is compared to the email

**Expected behavior**
The tool should compare corresponding values , i.e. have a case for username and email with different domains.

**Screenshots**
Screenshot 2022-12-05 at 14 10 10
Screenshot 2022-12-05 at 14 13 36

**Environment**
 - UST version: 2.7.5
 - OS type and version: Win Server 2012

Luci2015:
## Summary
* A fix to handle the remove/delete of accounts with different email than username (both having email formatted values), when the input is a csv list
* Modifies the behaviour so that when writing to csv file (--write-file) contains the mail value of the account instead of the username

## Testing Steps
* have in admin console an account with different email than username
* run the tool with  `--adobe-only-user-action write-file a.csv` argument
* look inside at the resulting .csv `username` field -> it should be the current account's email, that the API will use further for other action.
* run the tool with `--adobe-only-user-action delete --adobe-only-user-list a.csv` and the account should get deleted from admin console

Fixes #808

### DIFF
--- a/user_sync/engine/umapi.py
+++ b/user_sync/engine/umapi.py
@@ -1173,6 +1173,9 @@ class RuleProcessor(object):
         for umapi_name in self.stray_key_map:
             for user_key in self.get_stray_keys(umapi_name):
                 id_type, username, domain = self.parse_user_key(user_key)
+                # switch username with email if override in place:
+                if '@' in username and username in self.email_override.keys():
+                    username = self.email_override[username]
                 umapi = umapi_name if umapi_name else ""
                 if secondary_count:
                     row_dict = {'type': id_type, 'username': username, 'domain': domain, 'umapi': umapi}


adorton-adobe:
Thank you for this but I've decided to take a different approach. Since the central issue is that the email address is needed to identify the user for the UMAPI call, I've modified the CSV write function to write the email address instead of the username. The read function now checks for email primarily and only uses username if email is not provided.

As a consequence of this change, I updated the user key format to include email address.

The main changes are here - https://github.com/adobe-apiplatform/user-sync.py/commit/651d2117150d59030276297867eeef2d6c09093b

I've also updated all tests that were impacted by the key format change.

PR #811, opened 2023-03-02
User update fix

### LINKED ISSUE #810: Email address is not updating after 2.7.6 update.
**Description:**  Email address is not updating after 2.7.6 update. The User Sync Tool has started throwing errors that it can't update the email attribute of users with the following error: "code: "error.user.nonexistent" message: "User Id does not exist: xxxxxxxx@uvu.edu"".

**Steps to reproduce**
Customer indicates,  Students have accounts in the format “[idnumber@uvu.edu](mailto:idnumber@uvu.edu)”. When they are hired as employees, their primary email alias is updated to the format “[firstname.lastname@uvu.edu](mailto:firstname.lastname@uvu.edu)”. For the past 2 years, the UST has been correctly updating the email address. They just updated from 3.7.5 to 3.7.6 and now this issue has occurred.

**Testing**
Customer has tried updating to 2.8.0 and the issue still persists, but after rolling back to 2.7.5, all the errors disappeared. They have checked [idnumber@uvu.edu](mailto:idnumber@uvu.edu) and the email was successfully updated to [firstname.lastname@uvu.edu](mailto:firstname.lastname@uvu.edu).

Logs indicate this is the case.

Customer believes the fix https://github.com/adobe-apiplatform/user-sync.py/issues/808 had an unintended side effect.

**Environment**
UST version: 2.7.6
OS type and version: Win Server 2012

adorton-adobe:
## Summary

This is to fix a regression found in `v2.8.0` which prevents the sync tool from updating a user's email address. This was likely caused by the introduction of the email address to the static user key.

The fix I propose is to stop indexing user information using this static composite key. It never worked super well because of how the UMAPI and admin console identify users by email address under most circumstances but also require username in others.

I've created a system that indexes user info by multiple keys, but only one key is required to identify a user when looking it up in the index. This is implemented in a new class in the UMAPI engine file called `MultiIndex`.

More information can be found in the class's docstring:

```
    This data structure replaces the old convention of caching users in a
    dictionary indexed by a static composite key. The MultiIndex structure
    consists of a simple list (self.data) consisting of one or more dictionaries
    that follow a regular structure (e.g. a list of directory users or UMAPI
    users).

    This list is indexed by one or more keys. Each key should point to one
    record in self.data.

    When a record is fetched from the index, a record is returned if at least
    one key matches a record. This allows partial matches - i.e. when retrieving
    information for a user where the email address matches a record but the
    username does not (or vice versa).

    Example:

    >>> data = [{"key1": "foo", "key2": "bar", "other": "data"}]
    >>> mi = MultiIndex(data=data, key_names=["key1", "key2"])
    >>> mi.get(key1="foo", key2="bar")
    {"key1": "foo", "key2": "bar", "other": "data"}
    >>> mi.get(key1="foo", key2="invalid")
    {"key1": "foo", "key2": "bar", "other": "data"}
    >>> mi.get(key1="invalid", key2="invalid")
    None

    MultiIndex supports the addition of new individual records and
    the ability to update existing records. It does not support
    deletion because that would require a full reindex for each
    deletion.
```

## Testing Steps

I've tested this with one UMAPI target, primary + secondary and secondary only (which produces the expected UMAPI errors). I've also tested the update with Adobe-only CSV files (read/write).

Fixes #810

### DIFF
--- a/tests/test_umapi_engine.py
+++ b/tests/test_umapi_engine.py
@@ -9,7 +9,7 @@ from mock import MagicMock
 from tests.util import compare_iter
 from user_sync.connector.connector_umapi import Commands
 from user_sync.engine.common import AdobeGroup
-from user_sync.engine.umapi import UmapiTargetInfo, UmapiConnectors, RuleProcessor
+from user_sync.engine.umapi import UmapiTargetInfo, UmapiConnectors, RuleProcessor, MultiIndex
 
 
 @pytest.fixture
@@ -90,7 +90,7 @@ def test_create_umapi_user(rule_processor, mock_dir_user, mock_umapi_info):
     rp.logger.progress = progress_func
 
     key = rp.get_user_key(user['identity_type'], user['username'], user['domain'])
-    rp.directory_user_by_user_key[key] = user
+    rp.directory_user_index = MultiIndex(data=[user], key_names=['email', 'username'])
     rp.options['process_groups'] = True
     rp.push_umapi = True
 
@@ -132,7 +132,7 @@ def test_update_umapi_user(rule_processor, mock_dir_user, mock_umapi_user):
         conn = MockUmapiConnector()
         info = UmapiTargetInfo(None)
         user_key = rp.get_user_key(up_user['identity_type'], up_user['username'], up_user['domain'])
-        rp.directory_user_by_user_key[user_key] = up_user
+        rp.directory_user_index = MultiIndex(data=[up_user], key_names=['email', 'username'])
         commands = [rp.update_umapi_user(info, user_key, up_attrs, group_add, group_rem, mock_umapi_user)]
         rp.execute_commands(commands, conn)
         assert user_key in rp.updated_user_keys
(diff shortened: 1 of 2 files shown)


Luci2015 (commented):
new line 960 checks:
`if directory_user is None:`
at this stage we know that an umapi `user_key` was resolved from the list of `directory_users`.
To solve the fact that the username can't be updated in this version we could do at this stage the check if:
`normalize_string(umapi_user['username'])` != `normalize_string(directory_user['username'])` 
and log a warning under the `self.logger.debug("Adobe user matched on customer side: %s", user_key)`
`self.logger.warning('Username differences found: %s / %s',umapi_user['username'], directory_user['username'])`
This bit of code can be removed after the username updates will be possible. Sounds OK?

adorton-adobe:
Since we agreed to plan to release the username update feature in the same release as this fix, I don't think we need the temporary warning. 

Here's the feature branch for username update: https://github.com/adobe-apiplatform/user-sync.py/tree/feature/username-update

Once I get a chance to test it more I'll work on the configuration feature to select which fields to update.

adorton-adobe:
@Luci2015 - are you still testing? I hoped to have this merged a week ago. What can I help you with to get this finished up?

Luci2015 (approved):
works with csv input

Luci2015 (approved):

PR #863, opened 2025-03-21
Fix for 862

### LINKED ISSUE #862: ValueError: too many values to unpack (expected 4)
**Description**
When loading users from directory, some may have a comma in their username or email.
This causes the tool to throw:
"ValueError: too many values to unpack (expected 4)"
in the get_from_index function when unpacking the user_key tuple.

The fix is to check the input for email and username values, but the issue is that the error does not specify which user is the one having the issue.
Need a more verbose error.

**Environment**
 - UST version:  all
 - OS type and version: all

Luci2015:
## Summary
 This should fix the "Too many to unpack" error when the user's email or username contains a comma.

Fixes #862


adorton-adobe:
I would prefer an approach that does the error checking in the `parse_user_key()` method. I will draft a version of that fix.

